PR #8, opened 2021-04-09
Migrate original `auto-changelog.js` script

Gudahtt:
This script has been migrated from the [`metamask-extension` repo](https://github.com/MetaMask/metamask-extension/blob/0dbd678f393ae141d207e35b0132e2ee8cb118e9/development/auto-changelog.js). Functionally it should behave almost identically. The differences are:

* It now gets the current version from [the `npm_package_version` environment variable](https://docs.npmjs.com/cli/v6/using-npm/scripts#packagejson-vars) instead of from the hard-coded path to the MetaMask extension manifest.
* The `CHANGELOG.md` file is now read from the current directory where the script is run, rather than from the hard-coded path to the `CHANGELOG.md` file in this repository.
* The repository URL is taken from the `npm_package_repository_url` environment variable instead of being hard-coded.

These changes can all be seen in `src/cli.js`, which is the equivalent to the top-level script in the `metamask-extension` repo, `auto-changelog.js`.

These things will all be made more configurable in future PRs.

A mostly-empty `CHANGELOG.md` file has been added as well. It doesn't quite work with the `update-changelog` script yet, because it wasn't meant to accommodate packages with no releases. But it was enough to make it not completely blow up.

One silly test has been added to allow CI to pass without removing tests altogether. This will be improved in future PRs.

The package has one exported function, `updateChangelog`, to allow the changelog to be updated programmatically.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -1,0 +1,80 @@
+#!/usr/bin/env node
+/* eslint-disable node/no-process-exit */
+
+const fs = require('fs').promises;
+
+const { updateChangelog } = require('./updateChangelog');
+const { unreleased } = require('./constants');
+
+const command = 'yarn update-changelog';
+
+const helpText = `Usage: ${command} [--rc] [-h|--help]
+Update CHANGELOG.md with any changes made since the most recent release.
+
+Options:
+      --rc    Add new changes to the current release header, rather than to the
+                '${unreleased}' section.
+  -h, --help  Display this help and exit.
+
+New commits will be added to the "${unreleased}" section (or to the section for the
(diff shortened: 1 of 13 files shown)


src/cli.js
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env node
```
Gudahtt, line 1: To help with review, I created this branch in the extension to demonstrate all of the changes made to this file: https://github.com/MetaMask/metamask-extension/compare/auto-changelog-migration-changes

The other files migrated are all unchanged.

rekmarks (approved):
LGTM! All of the changes relative to the original make sense.

PR #23, opened 2021-04-12
Allow updating changelog when there are no tags

### LINKED ISSUE #13: Allow updating a changelog with no versions
The generated changelog is currently broken if there are no released versions.

Gudahtt:
The changelog can now be updated even when there are no tags. Previously it would break in `updateChangelog` when there was no most recent tag, and if you got past that it would put `vundefined` in the link definition for the "Unreleased" header. Now it is tolerant of there being no tags, and the repo URL is used for the link definition of the "Unreleased" header if there is no release to compare to.

Fixes #13


brad-decker (approved):
LGTM

PR #26, opened 2021-04-20
Move `update` to sub-command

Gudahtt:
The "update" action has been moved to a sub-command, in preparation for adding more commands in later PRs. Specifying a command has been made mandatory.

The `update-changelog` npm script has been renamed to `changelog`, to reflect that it will soon be used for more than just updates.

PR #28, opened 2021-04-20
Add `validate` command

### LINKED ISSUE #11: Expose API method and CLI command for validating changelog
We should explose a CLI command and an API method for validation the changelog. Currently the changelog is validated before it is updated, but we don't have a command yet to just validate.

Gudahtt:
The validate command will validate the formatting of the changelog. If the `--rc` flag is used, it will also ensure that the current version is included as a release header, and that there are no unreleased changes.

The CLI command displays a rudimentary diff if formatting problems are detected. This can be improved later with colours, and with highlighting within each line to show what has changed. It also doesn't yet highlight whitespace changes.

The validation logic is also exposed via the API as a function called `validateChangelog`. It doesn't include the fancy diff output, but it does throw an error with all of the required information for someone to construct the same output.

Closes #11

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -7,6 +7,11 @@ const yargs = require('yargs/yargs');
 const { hideBin } = require('yargs/helpers');
 
 const { updateChangelog } = require('./updateChangelog');
+const { generateDiff } = require('./generateDiff');
+const {
+  validateChangelog,
+  ChangelogFormattingError,
+} = require('./validateChangelog');
 const { unreleased } = require('./constants');
 
 const updateEpilog = `New commits will be added to the "${unreleased}" section (or \
@@ -16,6 +21,10 @@ changelog will be ignored.
 
 If the '--rc' flag is used and the section for the current release does not \
 yet exist, it will be created.`;
+
+const validateEpilog = `This does not ensure that the changelog is complete, \
+or that each change is in the correct section. It just ensures that the \
+formatting is correct. Verification of the contents is left for manual review.`;
 
 // eslint-disable-next-line node/no-process-env
 const npmPackageVersion = process.env.npm_package_version;
@@ -32,35 +41,106 @@ function isValidUrl(proposedUrl) {
   }
 }
 
+async function readChangelog(changelogFilename) {
+  return await fs.readFile(changelogFilename, {
+    encoding: 'utf8',
+  });
+}
+
+async function saveChangelog(changelogFilename, newChangelogContent) {
+  await fs.writeFile(changelogFilename, newChangelogContent);
+}
+
+async function update({
+  changelogFilename,
+  currentVersion,
+  isReleaseCandidate,
+  repoUrl,
+}) {
+  const changelogContent = await readChangelog(changelogFilename);
+
+  const newChangelogContent = await updateChangelog({
+    changelogContent,
+    currentVersion,
+    repoUrl,
+    isReleaseCandidate,
+  });
+
+  await saveChangelog(changelogFilename, newChangelogContent);
+  console.log('CHANGELOG updated');
+}
+
+async function validate({
+  changelogFilename,
+  currentVersion,
+  isReleaseCandidate,
+  repoUrl,
+}) {
+  const changelogContent = await readChangelog(changelogFilename);
+
+  try {
+    validateChangelog({
+      changelogContent,
+      currentVersion,
+      repoUrl,
+      isReleaseCandidate,
+    });
+  } catch (error) {
+    if (error instanceof ChangelogFormattingError) {
+      const { validChangelog, invalidChangelog } = error.data;
+      const diff = generateDiff(validChangelog, invalidChangelog);
+      console.error(`Changelog not well-formatted.\nDiff:\n${diff}`);
+      process.exit(1);
+    }
+    throw error;
+  }
+}
+
+function configureCommonCommandOptions(_yargs) {
+  return _yargs
+    .option('file', {
+      default: 'CHANGELOG.md',
+      description: 'The changelog file path',
+      type: 'string',
+    })
+    .option('currentVersion', {
+      default: npmPackageVersion,
+      description:
+        'The current version of the project that the changelog belongs to.',
+      type: 'string',
+    })
+    .option('repo', {
+      default: npmPackageRepositoryUrl,
+      description: `The GitHub repository URL`,
+      type: 'string',
+    });
+}
+
 async function main() {
   const { argv } = yargs(hideBin(process.argv))
     .command(
       'update',
       'Update CHANGELOG.md with any changes made since the most recent release.\nUsage: $0 update [options]',
       (_yargs) =>
-        _yargs
+        configureCommonCommandOptions(_yargs)
           .option('rc', {
             default: false,
             description: `Add new changes to the current release header, rather than to the '${unreleased}' section.`,
             type: 'boolean',
           })
-          .option('file', {
-            default: 'CHANGELOG.md',
-            description: 'The changelog file path',
-            type: 'string',
-          })
-          .option('currentVersion', {
-            default: npmPackageVersion,
-            description:
-              'The current version of the project that the changelog belongs to.',
-            type: 'string',
-          })
-          .option('repo', {
-            default: npmPackageRepositoryUrl,
-            description: `The GitHub repository URL`,
-            type: 'string',
-          })
           .epilog(updateEpilog),
+    )
+    .command(
+      'validate',
+      'Validate the changelog, ensuring that it is well-formatted.\nUsage: $0 validate [options]',
+      (_yargs) =>
+        configureCommonCommandOptions(_yargs)
+          .option('rc', {
+            default: false,
+            description: `Verify that the current version has a release header in the changelog`,
+            type: 'boolean',
+          })
+          .epilog(validateEpilog),
     )
     .strict()
     .demandCommand()
@@ -106,20 +186,21 @@ async function main() {
     process.exit(1);
   }
 
-  const changelogContent = await fs.readFile(changelogFilename, {
-    encoding: 'utf8',
-  });
-
-  const newChangelogContent = await updateChangelog({
-    changelogContent,
-    currentVersion,
-    repoUrl,
-    isReleaseCandidate,
-  });
-
-  await fs.writeFile(changelogFilename, newChangelogContent);
-
-  console.log('CHANGELOG updated');
+  if (argv._ && argv._[0] === 'update') {
+    await update({
+      changelogFilename,
+      currentVersion,
+      isReleaseCandidate,
+      repoUrl,
+    });
+  } else if (argv._ && argv._[0] === 'validate') {
+    await validate({
+      changelogFilename,
+      currentVersion,
+      isReleaseCandidate,
+      repoUrl,
+    });
+  }
 }
 
 main().catch((error) => {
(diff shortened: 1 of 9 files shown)


Gudahtt:
This depends upon #26

src/cli.js
```diff
@@ -16,30 +21,94 @@ changelog will be ignored.
...
           .epilog(updateEpilog),
     )
+    .command(
+      'validate',
+      'Validate the changelog, ensuring that it is well-formatted.\nUsage: $0 validate [options]',
+      (_yargs) =>
+        configureCommonCommandOptions(_yargs)
+          .option('rc', {
+            default: false,
+            description: `Verify that the current version has a release header in the changelog`,
```
Gudahtt, line 107: I left `rc` out of the "commonCommandOptions" group because it was helpful to have a separate description for this flag for each command, since it implies something a bit different in each case.

src/validateChangelog.js
```diff
@@ -0,0 +1,74 @@
...
+  const changelog = parseChangelog({ changelogContent, repoUrl });
+
+  // Ensure release header exists, if necessary
+  if (
+    isReleaseCandidate &&
+    !changelog
+      .getReleases()
+      .find((release) => release.version === currentVersion)
+  ) {
+    throw new Error(
```
Gudahtt, line 55: Not sure on whether I should bother with a custom Error class for these other two errors as well :thinking: 

I only did that for the formatting error because I wanted to return extra information.

Gudahtt: I suppose one alternative would be a Result-style return value, but we don't have any conventional types for that. Returning `true` for "valid" and an error object for "invalid" seemed bad because the invalid case is still truthy, making it easy to misuse that result. But returning `false` or `null` for the valid case also seemed semantically confusing.

Gudahtt: I'll update it for now to use a common Error type, so that the "invalid changelog" cases can be distinguished from totally unexpected errors.

Gudahtt: Done in 8b592edc2498f7f2a2f023f2ea0d5b5a7f750db8

src/validateChangelog.test.js
```diff
@@ -0,0 +1,445 @@
...
+    expect(() =>
+      validateChangelog({
+        changelogContent: branchingChangelog,
+        currentVersion: '1.0.0',
+        repoUrl:
+          'https://github.com/ExampleUsernameOrOrganization/ExampleRepository',
+        isReleaseCandidate: false,
+      }),
+    ).not.toThrow();
+  });
```
rekmarks, line 290: nit: I was surprised to find a success / `not.toThrow()` case here. It might be better to group them.

src/cli.js
```diff
@@ -17,6 +23,10 @@ changelog will be ignored.
 If the '--rc' flag is used and the section for the current release does not \
 yet exist, it will be created.`;
 
+const validateEpilog = `This does not ensure that the changelog is complete, \
+or that each change is in the correct section. It just ensures that the \
+formatting is correct. Verification of the contents is left for manual review.`;
```
rekmarks, line 28: Aside: I've been wondering with increasing frequency if backticks / template strings are superior to single/double-quoted strings and concatenation using `+`. I'm starting to think that may not be the case 🤔

src/cli.js
```diff
@@ -32,36 +42,110 @@ function isValidUrl(proposedUrl) {
...
+      changelogContent,
+      currentVersion,
+      repoUrl,
+      isReleaseCandidate,
+    });
+  } catch (error) {
+    if (error instanceof ChangelogFormattingError) {
+      const { validChangelog, invalidChangelog } = error.data;
+      const diff = generateDiff(validChangelog, invalidChangelog);
+      console.error(`Changelog not well-formatted.\nDiff:\n${diff}`);
```
rekmarks, line 93: nit: I think this would look better:

```suggestion
      console.error(`Changelog not well-formatted. Diff:\n\n${diff}`);
```

rekmarks (changes requested):
This was great to review. Well conceived structure, great tests, and neat use of diffing!

I don't have any blocking outright change requests, but rather questions and comments that should be addressed before merging.

src/cli.js
Gudahtt: Oh? I find them especially useful with longer strings. I like not having to use `\n` anywhere.

src/generateDiff.js
```diff
@@ -0,0 +1,41 @@
...
+ * @returns {string} The genereated text diff
+ */
+function generateDiff(before, after) {
+  const changes = diff.diffLines(before, after);
+  const diffLines = [];
+  const preceedingContext = [];
+  for (const { added, removed, value } of changes) {
+    const lines = value.split('\n');
+    // remove trailing newline
+    lines.pop();
```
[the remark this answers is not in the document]
Gudahtt: I didn't know actually. I meant to come back to this and forgot. This is a mistake I believe.

Because `diffLines` is meant to operate on lines, I had wrongly assumed that the newlines between each changeset would be removed. That is not the case though, so when I'm joining with `\n` below I'm introducing a _new_ newline that I've removed here to compensate.

I'll update this to just preserve the newline in the text.

src/generateDiff.js
```diff
@@ -0,0 +1,41 @@
...
+    } else {
+      // If a changed line has been included already, add up to 2 lines of context
+      if (diffLines.length) {
+        diffLines.push(...lines.slice(0, 2).map((line) => ` ${line}`));
+        lines.splice(0, 2);
+      }
+      // stash last 2 lines for context in case there is another change
+      if (lines.length) {
+        preceedingContext.push(...lines.slice(-2));
+      }
```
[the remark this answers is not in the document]
Gudahtt: Yes, I think you're right about that possibility. This is a bug.

I'm going to update this to simplify the logic by referencing all 3 changes each iteration (previous, current, next) to avoid having to cache anything.

src/cli.js
rekmarks: String style fight! String style fight! String style fight!

src/generateDiff.js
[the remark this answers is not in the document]
Gudahtt: Well, this ended up being much more complicated than anticipated! I tried and failed to fix this many times, namely because of the ambiguity introduced when files don't end in newlines.

I ended up writing comprehensive tests for `generateDiff`, and rewriting most of the diff logic. Result here: b1375e802173782feb4c6e2dccf98d3f081a2dbf

src/generateDiff.js
```diff
@@ -0,0 +1,41 @@
...
+ * respectively, and includes 2 lines of unchanged content around each changed
+ * section where possible.
+ * @param {string} before - The string representing the base for the comparison.
+ * @param {string} after - The string representing the changes being compared.
+ * @returns {string} The genereated text diff
+ */
+function generateDiff(before, after) {
+  const changes = diff.diffLines(before, after);
+  const diffLines = [];
+  const preceedingContext = [];
```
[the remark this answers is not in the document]
Gudahtt: I ended up removing this in b1375e802173782feb4c6e2dccf98d3f081a2dbf

src/validateChangelog.test.js
Gudahtt: I had thought these were ordered by feature, roughly? But looking at it now, that's not clear either. I'll try to order them a bit more sensibly.

Gudahtt: Tests re-ordered in https://github.com/MetaMask/auto-changelog/pull/28/commits/bcebb91c2ddcb6e7fcaf321472f8526a7dd8e502

src/generateDiff.js
```diff
@@ -11,30 +18,64 @@ const diff = require('diff');
...
-      }
-      diffLines.push(...lines.map((line) => `${added ? '+' : '-'}${line}`));
-    } else {
-      // If a changed line has been included already, add up to 2 lines of context
-      if (diffLines.length) {
-        diffLines.push(...lines.slice(0, 2).map((line) => ` ${line}`));
-        lines.splice(0, 2);
+  // `diffLines` will always return at least one change
+  const lastChange = changes[changes.length - 1];
+  const penultimateChange = changes[changes.length - 2] || {};
```
rekmarks, line 23: nit: My brain doesn't like that `lastChange` is declared before `penultimateChange`. Relative to the index it obviously does make sense, so deciding what to do about that is left as an exercise to the author 😛

src/generateDiff.js
```diff
@@ -11,30 +18,64 @@ const diff = require('diff');
...
-        diffLines.push(...preceedingContext);
-        preceedingContext.splice(0, preceedingContext.length);
-      }
-      diffLines.push(...lines.map((line) => `${added ? '+' : '-'}${line}`));
-    } else {
-      // If a changed line has been included already, add up to 2 lines of context
-      if (diffLines.length) {
-        diffLines.push(...lines.slice(0, 2).map((line) => ` ${line}`));
-        lines.splice(0, 2);
+  // `diffLines` will always return at least one change
```
rekmarks, line 21: Will it always return at least one _change_ or one _value_?

src/generateDiff.js
```diff
@@ -11,30 +18,64 @@ const diff = require('diff');
...
+      const changedLines = [];
+      if (added || removed) {
+        // Add up to 2 lines of context before each change
+        const previousContext = changes[index - 1];
+        if (
+          previousContext &&
+          !previousContext.added &&
+          !previousContext.removed
+        ) {
+          const hasPreviousChange = index > 1;
```
rekmarks, line 53: This could use some more explanation, either be a comment here or expanding the comment on L21 above.

src/generateDiff.test.js
```diff
@@ -0,0 +1,325 @@
+const _outdent = require('outdent');
+
+const outdent = _outdent({ trimTrailingNewline: false });
+const { generateDiff } = require('./generateDiff');
+
+const testCases = [
+  {
+    description: 'should return an empty string when comparing empty files',
+    before: '\n',
+    after: '\n',
```
rekmarks, line 10: Should these be actually empty (i.e. `''`) or is this as intended?

src/generateDiff.js
```diff
@@ -1,8 +1,15 @@
 const diff = require('diff');
 
+function getTrimmedLines(value) {
```
rekmarks, line 3: This could use a docstring to help the reader understand why this isn't simply:

```
return value.trim().split('\n')
```

rekmarks (changes requested):
Alright, one last round!

src/generateDiff.js
Gudahtt: Ah, right, I see that I was using the term 'change' to mean different things in this file :thinking:
I can re-word this to be a bit more clear. Maybe I can call this a `diffResult` rather than a "change"? As it includes both additions, removals, and sections where no change occured.

src/generateDiff.test.js
Gudahtt: This was intended - I was considering an "empty file" to be just a newline, as this is what an empty text file looks like on POSIX-like systems. All text files should end in newlines according to POSIX.

This just made for a nicer base test case, as the lack of newline notices really clutter things. Those are covered in other tests.

src/generateDiff.js
Gudahtt: The library calls these "change objects", where each "change object" might be a removal, an addition, or a section that hadn't changed.

Gudahtt:
Just pushed one more commit to improve variable names and comments. I think everything has been addressed - let me know if not!

rekmarks (approved):
LGTM!

PR #29, opened 2021-04-20
Fix release header sorting

Gudahtt:
The release header sorting was broken in two ways. First, the sort function was totally broken - it returned a boolean instead of the expected `-1` or `1`. Second, the release link generation would always consider the _numerical_ previous release to be the previous release for comparison purposes, even if it was released _after_.

The sorting function now correctly returns numbers, as `Array.prototype.sort` expects. The release link logic has been updated to ensure the "previous" release is both numerically and chronologically previous as well.

We don't yet validate that the releases are sorted chronologically, but that is the assumption for now. We can add validation using release dates at some point in the future.

### DIFF
--- a/src/changelog.js
+++ b/src/changelog.js
@@ -64,9 +64,12 @@ function getTagUrl(repoUrl, tag) {
 }
 
 function stringifyLinkReferenceDefinitions(repoUrl, releases) {
-  const orderedReleases = releases
+  const releasesOrderedByVersion = releases
     .map(({ version }) => version)
-    .sort((a, b) => semver.gt(a, b));
+    .sort((a, b) => {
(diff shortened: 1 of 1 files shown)


Gudahtt:
These bugs were discovered while working on unit tests for #28. See the `branchingChangelog` fixture in `validateChangelog.test.js` for an example changelog that exposes this problem.

Gudahtt:
The change in logic here is a bit... nuanced, but the intent was explained in some detail in the inline comments. Please let me know if something is still unclear!

brad-decker (approved):
LGTM. The change in logic matches the comments and, at least to me, was not difficult to understand.

PR #30, opened 2021-04-20
Allow configuring the changelog file path

### LINKED ISSUE #9: Allow configuring changelog file path
The changelog is assumed to be at the path `CHANGELOG.md` in the current directory. We should allow updating changelogs at a custom path, with any filename.

Gudahtt:
The `update` CLI command now allows configuring the changelog filepath.

Additionally, validation has been added to improve the error message if the file doesn't exist or isn't writable. Previously this would have already failed, but with a more confusing message.

Closes #9

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 /* eslint-disable node/no-process-exit */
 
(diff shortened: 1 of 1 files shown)


brad-decker (approved):
LGTM

PR #31, opened 2021-04-20
Allow configuring 'currentVersion'

### LINKED ISSUE #10: Allow setting version via CLI
The version is currently taken from the `npm_package_version` environment variable. We should allow the version to be passed in as a CLI argument as well, to cover any case where this environment variable isn't set or doesn't have the correct version.

Gudahtt:
Rather than always obtaining the 'currentVersion' of the project from an environment variable, it can now be passed in via a CLI flag instead.

Additionally, the validation for the version has been improved. It will now throw if the current version isn't a valid SemVer version, as our update logic depends upon it being valid. We also now ensure that the error message for a messing version is only printed when it was strictly required (it wasn't used for non-RC updates), and that the script terminates at that point.

Closes #10

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -2,6 +2,7 @@
 /* eslint-disable node/no-process-exit */
 
 const { promises: fs, constants: fsConstants } = require('fs');
+const semver = require('semver');
 const yargs = require('yargs/yargs');
 const { hideBin } = require('yargs/helpers');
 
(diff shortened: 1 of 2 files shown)


brad-decker (approved):
💯 LGTM.

Gudahtt:
Rebased to resolve conflicts

brad-decker (approved):

PR #41, opened 2021-05-03
Add `root` parameter

### LINKED ISSUE #34: Monorepo support
This script should support monorepos. This might be as simple as adding an option to specify the repository root.

Gudahtt:
The `root` parameter allows specifying the root project directory. This is used to resolve the changelog file path (unless an absolute path is specified), and it's used by the `update` command to ensure the search for changes since the last release only includes changes in the given root directory.

This new option is exposed via the API only on the `update` function, with the `rootProjectDirectory` option.

This was the last feature required for monorepo support.

Closes #34

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -2,6 +2,7 @@
 /* eslint-disable node/no-process-exit */
 
 const { promises: fs, constants: fsConstants } = require('fs');
+const path = require('path');
 const semver = require('semver');
 const yargs = require('yargs/yargs');
 const { hideBin } = require('yargs/helpers');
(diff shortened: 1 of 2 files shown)


src/cli.js
```diff
@@ -178,28 +189,58 @@ async function main() {
...
   }
 
+  if (projectRootDirectory) {
+    try {
+      const stat = await fs.stat(projectRootDirectory);
+      if (!stat.isDirectory()) {
+        console.error(
+          `Project root must be a directory: '${projectRootDirectory}'`,
+        );
+        process.exit(1);
```
rekmarks, line 199: This is probably out of scope for this PR, but I recently realized how strongly Node advises against the log-and-process.exit pattern: https://nodejs.org/api/process.html#process_process_exit_code

Going through `snaps-cli` and getting rid of 17 cases of the same is on my TODO-list 😑

rekmarks (approved):
LGTM!

src/cli.js
Gudahtt: TIL. Easy enough to avoid this pattern in this case at least. I can fix this in a follow-up PR.

Gudahtt:
Found a couple of silly mistakes in further testing - see the newest two commits.

rekmarks (approved):

src/cli.js
Gudahtt: Resolved here: https://github.com/MetaMask/auto-changelog/pull/43

PR #43, opened 2021-05-04
Avoid `process.exit`

Gudahtt:
We now set `process.exitCode` in the event of a CLI error, rather than calling `process.exit` directly with the desired exit code. This is in accordance with [recommendations in the Node.js documentation](https://nodejs.org/api/process.html#process_process_exit_code).


rekmarks (approved):
LGTM!

PR #48, opened 2021-05-04
Fix Unreleased changes detection

Gudahtt:
The command `auto-changelog validate --rc` would always fail claiming that there were Unreleased changes, even when there were none. This has been fixed, and a test has been added to capture this case.


rekmarks (approved):
LGTM!

PR #53, opened 2021-05-05
Fix mistaking newline between categories as part of change entry

Gudahtt:
Our `parseChangelog` function was mistakenly interpreting newlines between change categories as part of the change entry. Now it does not.

A new test has been added for this fix, along with a second one to verify that we're treating the newline between releases appropriately as well (this was working already but wasn't tested).


rekmarks (approved):
🛠️

PR #59, opened 2021-05-17
Migrate to TypeScript

rekmarks:
Migrates the whole project to TypeScript. Care has been taken to change as little as possible, but in some places I had to add behavior-altering code to please `tsc`.

At the time of writing, there are a handful of type-related, outstanding TODOs, which I've highlighted with inline comments.


.eslintrc.js
```diff
@@ -5,10 +5,19 @@ module.exports = {
 
   overrides: [
     {
-      files: ['*.test.js'],
+      files: ['*.ts'],
+      extends: ['@metamask/eslint-config-typescript'],
+      rules: {
+        'no-shadow': 'off',
+        '@typescript-eslint/no-shadow': 'error',
+        'node/prefer-global/url': 'off',
```
rekmarks, line 13: Installing `@types/node` didn't work to get `tsc` to stop complaining about the one use of the `URL` global we have. Not sure what to do here.

src/cli.ts
```diff
@@ -42,30 +44,39 @@ function isValidUrl(proposedUrl) {
...
   changelogPath,
   currentVersion,
   isReleaseCandidate,
   repoUrl,
   projectRootDirectory,
-}) {
+}: UpdateOptions) {
   const changelogContent = await readChangelog(changelogPath);
 
+  // TODO: This can be undefined. Is that OK?
```
rekmarks, line 79: `updateChangelog` can return `undefined`. Should we save that content?

src/cli.ts
```diff
@@ -248,7 +269,8 @@ async function main() {
   } else if (argv._ && argv._[0] === 'validate') {
     await validate({
       changelogPath,
-      currentVersion,
+      // TODO: Is it OK to assign an empty string here?
+      currentVersion: currentVersion || '',
```
rekmarks, line 273: This seems bad. Should we add some input validation here?

src/parseChangelog.ts
```diff
@@ -95,10 +111,13 @@ function parseChangelog({ changelogContent, repoUrl }) {
         throw new Error(`Malformed category header: '${truncated(line)}'`);
       }
       const isFirstCategory = mostRecentCategory === null;
-      finalizePreviousChange({ removeTrailingNewline: !isFirstCategory });
-      mostRecentCategory = results[1];
+      finalizePreviousChange({
+        removeTrailingNewline: !isFirstCategory,
+      });
+      // TODO: Get rid of typecast?
+      mostRecentCategory = results[1] as changeCategories;
```
rekmarks, line 118: Do we **know** that `results[1]` will be a valid change category?

src/updateChangelog.ts
```diff
@@ -174,11 +189,13 @@ async function updateChangelog({
...
   ) {
-    changelog.addRelease({ version: currentVersion });
+    // TODO: Get rid of typecast?
+    changelog.addRelease({ version: currentVersion as string });
   }
 
   if (isReleaseCandidate && hasUnreleasedChanges) {
-    changelog.migrateUnreleasedChangesToRelease(currentVersion);
+    // TODO: Get rid of typecast?
+    changelog.migrateUnreleasedChangesToRelease(currentVersion as string);
```
rekmarks, line 198: `currentVersion` can be omitted, making theses cases somewhat ambiguous.

.eslintrc.js
Gudahtt: Welp. Found this issue: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34960

That's unfortunate. Disabling this rule and using the import seems like a reasonable workaround for now :+1:

.github/workflows/lint-test.yml
```diff
@@ -1,30 +1,31 @@
...
-        node-version: ${{ matrix.node-version }}
-    - run: yarn --frozen-lockfile
-    - run: yarn lint
-    - run: yarn test
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: yarn --ignore-scripts --frozen-lockfile
```
Gudahtt, line 21: Probably better to omit `--ignore-scripts` until after we add `@lavamoat/allow-scripts`. I can do that in a separate PR
```suggestion
      - run: yarn --frozen-lockfile
```

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Changelog
+
```
Gudahtt, line 2: The changes in this file are disallowed by `auto-changelog` (they follow the example on KeepAChangelog.com)

package.json
```diff
@@ -24,9 +23,10 @@
     "test:watch": "jest --watch",
     "prepublishOnly": "yarn lint && yarn test",
     "lint:eslint": "eslint . --cache --ext js,ts",
-    "lint:json": "prettier '**/*.json' --ignore-path .gitignore",
-    "lint": "yarn lint:eslint && yarn lint:json --check",
-    "lint:fix": "yarn lint:eslint --fix && yarn lint:json --write",
+    "lint:misc": "prettier '**/*.json' '**/*.md' '**/*.yml' --single-quote --ignore-path .gitignore",
```
Gudahtt, line 26: This should be updated to omit `CHANGELOG.md`, which doesn't follow the same formatting rules

Gudahtt: Also, should we add a `.prettierrc` instead of specifying `--single-quote` here via the CLI?

src/constants.ts
```diff
@@ -21,34 +20,30 @@
...
-const changeCategories = {
-  Added: 'Added',
-  Changed: 'Changed',
-  Deprecated: 'Deprecated',
-  Fixed: 'Fixed',
-  Removed: 'Removed',
-  Security: 'Security',
-  Uncategorized: 'Uncategorized',
-};
+export enum changeCategories {
```
Gudahtt, line 23: Could we make this singular instead? I found this rather confusing.

Previously `changeCategories` made sense because it was a collection. But now that it's an enum, it's not really a collection anymore. Any use of this enum does refer to a _single_ category.

For example: ```type ReleaseChanges = Record<changeCategories, string[]>;```
Here we're mapping a _single category_ to an array of strings, not a collection of categories to an array of strings.

src/changelog.ts
```diff
@@ -21,7 +56,11 @@ function stringifyCategory(category, changes) {
   return `${categoryHeader}\n${changeDescriptions}`;
 }
 
-function stringifyRelease(version, categories, { date, status } = {}) {
+function stringifyRelease(
+  version: string,
```
Gudahtt, line 60: This might be a more descriptive type to use here:
```suggestion
  version: Version | typeof unreleased,
```

src/changelog.ts
```diff
@@ -51,22 +93,25 @@ function stringifyReleases(releases, changes) {
...
 
-function stringifyLinkReferenceDefinitions(repoUrl, releases) {
+function stringifyLinkReferenceDefinitions(
+  repoUrl: string,
+  releases: ReleaseMetadata[],
+) {
   const releasesOrderedByVersion = releases
     .map(({ version }) => version)
-    .sort((a, b) => {
+    .sort((a: string | semver.SemVer, b: string | semver.SemVer) => {
```
Gudahtt, line 114: The type for these values should be `Version` I think? I don't follow where `semver.SemVer` is coming from

```suggestion
    .sort((a: Version, b: Version) => {
```

src/changelog.ts
```diff
@@ -100,7 +145,7 @@ function stringifyLinkReferenceDefinitions(repoUrl, releases) {
       const versionIndex = orderedReleases.indexOf(version);
       const previousVersion = orderedReleases
         .slice(versionIndex)
-        .find((releaseVersion) => {
+        .find((releaseVersion: string | semver.SemVer) => {
```
Gudahtt, line 148: Same thing here
```suggestion
        .find((releaseVersion: Version) => {
```

src/changelog.ts
```diff
@@ -1,16 +1,51 @@
...
+  /**
+   * The status of the release (e.g. 'WITHDRAWN', 'DEPRECATED')
+   */
+  status?: string;
+}
+
+/**
+ * Release changes, organized by category.
+ */
+type ReleaseChanges = Record<changeCategories, string[]>;
```
Gudahtt, line 37: The `ReleaseChanges` objects don't contain _all_ categories, so I think a `Partial` is warranted here

```suggestion
type ReleaseChanges = Partial<Record<changeCategories, string[]>>;
```

This allows getting rid of the couple of `as ReleaseChanges` later on, though not without some cost (TypeScript needs a little extra help to know that we're handling absent categories properly).

src/changelog.ts
```diff
@@ -283,7 +324,7 @@ class Changelog {
    * @param {Version} version - The version of the release being retrieved.
    * @returns {ReleaseChanges} The changes included in the given released.
    */
-  getReleaseChanges(version) {
+  getReleaseChanges(version: string) {
```
Gudahtt, line 327: This should be `Version` as well
```suggestion
  getReleaseChanges(version: Version) {
```

src/changelog.ts
```diff
@@ -249,15 +288,17 @@ class Changelog {
    * @param {Version} version - The release version to migrate unreleased
    *   changes to.
    */
-  migrateUnreleasedChangesToRelease(version) {
+  migrateUnreleasedChangesToRelease(version: string) {
```
Gudahtt, line 291:
```suggestion
  migrateUnreleasedChangesToRelease(version: Version) {
```

Gudahtt (commented):
Not quite through it all yet, but it looks good so far!

I did notice that types remain in most of the JSDoc comments. Now that we have real types, those are redundant and should probably be removed.

src/constants.ts
rekmarks: Done in: 8323b60

package.json
rekmarks: Done in: 47b3868

.github/workflows/lint-test.yml
rekmarks: Done in: 6025b58

CHANGELOG.md
rekmarks: Reverted in: 47b3868

However, I have thoughts about this: #60

rekmarks:
@Gudahtt re: JSDoc type annotations, I did a pass removing all of them (and updating some links for VS Code compatibility) in 69aae24. Note that TSDoc / TypeDoc / Whatever VS code uses is not compatible with `@throws`, so I removed those tags completely.

src/changelog.ts
rekmarks: > ~a little~ _an almost unbelievable amount of_ extra help

FTFY 😅 

Not sure how to do this in a sane way without adding a bunch of postfix non-null assertions (and disabling the related ESLint rule). I'm wondering if there's even a bug in TypeScript. Like, it thinks that `changeDescriptions` in this snippet is `(string | undefined)[]`:

https://github.com/MetaMask/auto-changelog/blob/bcb7beade10c39d57a287feeed75b83ac3a87e83/src/updateChangelog.ts#L72-L83

Gudahtt: That type seems correct to me. `Partial` _does_ imply that the key can be present and assigned `undefined`. Or it implies that it's possible anyway - either complete absence or an `undefined` assignment. I guess there's no way to distinguish those things with types :thinking:

src/changelog.ts
```diff
@@ -246,22 +286,22 @@ class Changelog {
...
 
-    for (const category of Object.keys(unreleasedChanges)) {
-      if (releaseChanges[category]) {
+    for (const category of Object.keys(unreleasedChanges) as ChangeCategory[]) {
+      if (category in releaseChanges) {
         releaseChanges[category] = [
-          ...unreleasedChanges[category],
-          ...releaseChanges[category],
+          ...(unreleasedChanges[category] as string[]),
+          ...(releaseChanges[category] as string[]),
```
Gudahtt, line 304: The issue with this cast seems to be that TypeScript isn't good at inferring types of object properties.

Line 301 had been `if (releaseChanges[category])` beforehand, which seems a bit better than what's there now. But it still didn't trigger the type inference to rule out `undefined`. I was able to get it to work by pulling it out into a separate variable. e.g.
```
const releaseChanges = releaseChangeCategories[category];
if (releaseChanges) {
  releaseChanges.unshift(...(unreleasedChanges as string[]));
} else {
  releaseChangeCategories[category] = unreleasedChanges;
}
```

Which... isn't great, but demonstrates what I'm talking about at least. Oh also I renamed the variable `releaseChanges` to `releaseChangeCategories` so I could use the name `releaseChanges` for a specific category, and likewise for the unreleased changes.

src/changelog.ts
Gudahtt: The "TypeScript insanity" changes you made seem pretty reasonable overall. Some of the non-null assertions/casting could be avoided by adding type guards, but it's awkward in literally all of those cases I think. Maybe worth trying anyway, even if it is awkward? But certainly nothing I'd block on.

src/cli.ts
```diff
@@ -1,19 +1,21 @@
-#!/usr/bin/env node
```
Gudahtt, line 1: :eyes: This shebang is a strict requirement for this to function as a CLI. Is there any way to preserve this?

src/cli.ts
Gudahtt: Ah, good catch! This was a mistake. We should not save anything in this case.

src/cli.ts
Gudahtt: See this comment in `updateChangelog`:
>@param options.currentVersion - The current version. Required if `isReleaseCandidate` is set, but optional otherwise.

I believe it was my intention to consider it "conditionally required" in the same way for `validate` as well, but I forgot to explain this in the doc comment.

Gudahtt: This would mean making this optional in `ValidateOptions` as well

src/cli.ts
rekmarks: Hm, it'll still function as a CLI if it's invoked as `node path/to/script`, but obviously not if invoked directly as an executable, i.e. `./path/to/script`.

I could add an override in the ESLint config. That's what I've done in the past. Also, if `cli.js` is intended to be the CLI entry point, should we use the [`bin` field in package.json](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#bin) for that?

Gudahtt: We are already using the `bin` field.

Having to invoke this using `node` would be very inconvenient, and would break CI in the extension for example.

Which rule is preventing this from being used? :thinking:

rekmarks: I just noticed we were already using the `bin` field as you replied 😓 

It's `node/shebang` that prevents it, because `cli.ts` is not the file that the package.json `bin` field is pointing to, I'm guessing. Either way, it's restored in 4a08eed, and I confirmed the shebang is there in `dist/cli.js` as well.

src/cli.ts
```diff
@@ -177,7 +195,10 @@ async function main() {
     root: projectRootDirectory,
   } = argv;
 
-  if (isReleaseCandidate && !currentVersion) {
+  if (
+    isReleaseCandidate &&
+    (!currentVersion || typeof currentVersion !== 'string')
```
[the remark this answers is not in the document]
rekmarks: I couldn't tell you if my life depended on it. Removed in: 84406e0

src/cli.ts
Gudahtt: Ahhh I see, makes sense :+1:

src/cli.ts
rekmarks: Done in: 203c240

src/generateDiff.ts
```diff
@@ -1,12 +1,14 @@
-const diff = require('diff');
+import * as diff from 'diff';
+
+export type Change = diff.Change & { noNewline?: boolean };
```
Gudahtt, line 3: Does this need to be exported? :thinking: I think it's strictly internal.

```suggestion
type Change = diff.Change & { noNewline?: boolean };
```

src/changelog.ts
rekmarks: https://github.com/microsoft/TypeScript/pull/43947 is exactly what we need here! In lieu of that, we'll have to go with one ugly or the other.

I tried type guards locally yesterday and I didn't like it any better. I think we might as well stick with this until TypeScript 5 (I guess, since they categorized the above PR as breaking).

src/runCommand.ts
```diff
@@ -11,19 +12,26 @@ const spawn = require('cross-spawn');
...
+  args?: readonly string[],
+) {
+  const output: string[] = [];
   let mostRecentError;
   let errorSignal;
   let errorCode;
   const internalError = new Error('Internal');
   try {
-    await new Promise((resolve, reject) => {
-      const childProcess = spawn(command, args, { encoding: 'utf8' });
```
Gudahtt, line 26: Whoops! Not sure how that `encoding` option snuck in there. That seems to not exist.

src/changelog.ts
rekmarks: I restored `if (releaseChanges[category])`, but otherwise left things as they are. I prefer spreading over `.unshift`, and `.unshift` didn't get rid of _all_ casts anyway. 3cc4e79

src/runCommand.ts
```diff
@@ -66,7 +76,7 @@ async function runCommand(command, args) {
       } else {
         errorMessage = `Exited with code '${errorCode}'`;
       }
-      const improvedError = new Error(errorMessage);
+      const improvedError: Error & { cause?: string } = new Error(errorMessage);
```
Gudahtt, line 79: `cause` should be an error:
```suggestion
      const improvedError: Error & { cause?: Error } = new Error(errorMessage);
```

src/updateChangelog.ts
```diff
@@ -91,44 +99,51 @@ function getAllLoggedPrNumbers(changelog) {
...
+ * @param options
+ * @param options.changelogContent - The current changelog
+ * @param options.currentVersion - The current version. Required if
+ * `isReleaseCandidate` is set, but optional otherwise.
+ * @param options.repoUrl - The GitHub repository URL for the current project.
+ * project.
+ * @param options.isReleaseCandidate - Denotes whether the current project.
+ * is in the midst of release preparation or not. If this is set, any new
+ * changes are listed under the current release header. Otherwise, they are
+ * are listed under the 'Unreleased' section.
```
Gudahtt, line 133: `are` repeated twice:
```suggestion
 * listed under the 'Unreleased' section.
```

src/cli.ts
rekmarks: Fixed in 47e9a7a

src/updateChangelog.ts
```diff
@@ -91,44 +99,51 @@ function getAllLoggedPrNumbers(changelog) {
...
- *   used to filter results from various git commands. This path is assumed to
- *   be either absolute, or relative to the current directory. Defaults to the
- *   root of the current git repository.
- * @returns {string} The updated changelog text
+ * @param options
+ * @param options.changelogContent - The current changelog
+ * @param options.currentVersion - The current version. Required if
+ * `isReleaseCandidate` is set, but optional otherwise.
+ * @param options.repoUrl - The GitHub repository URL for the current project.
+ * project.
```
Gudahtt, line 129: The word `project` was moved up - this line isn't needed

```suggestion
```

src/updateChangelog.ts
```diff
@@ -156,10 +171,11 @@ async function updateChangelog({
 
   const loggedPrNumbers = getAllLoggedPrNumbers(changelog);
   const newCommits = commits.filter(
-    ({ prNumber }) => !loggedPrNumbers.includes(prNumber),
+    ({ prNumber }) => !loggedPrNumbers.includes(prNumber as any),
```
Gudahtt, line 174: Instead of using `as any` here, we could satisfy `tsc` by adding `prNumber === undefined || ` to the front of this expression

src/changelog.ts
Gudahtt: It's in the 4.4 milestone at the moment. TypeScript doesn't follow semver - they often ship breaking changes in minor releases.

So... maybe relatively soon? :crossed_fingers:

src/updateChangelog.ts
Gudahtt: Now that you'd added error handling above, this is guaranteed to be a string, despite `tsc` still not recognizing this.

So maybe these casts are fine? I don't know offhand how to reorganize this to avoid it, other than by duplicating that type guard.

For both of these casts, we'd might as well cast it to `Version` though.

src/updateChangelog.ts
rekmarks:
```suggestion
    ({ prNumber }) => prNumber === undefined || !loggedPrNumbers.includes(prNumber),
```

src/validateChangelog.ts
```diff
@@ -0,0 +1,117 @@
...
+  constructor(currentVersion: Version) {
+    super(`Current version missing from changelog: '${currentVersion}'`);
+  }
+}
+
+/**
+ * Represents a formatting error in a changelog.
+ */
+export class ChangelogFormattingError extends InvalidChangelogError {
+  public data: Record<string, string>;
```
Gudahtt, line 34: Curious that in `cli.ts`, when this is referenced, it's able to infer that it has entries for `validChangelog` and `invalidChangelog` :thinking: I don't quite follow how it knows those properties exist. Some kinda fancy type inference maybe.

src/validateChangelog.js
```diff
@@ -1,112 +0,0 @@
...
- * @param {string} options.changelogContent - The current changelog
- * @param {Version} options.currentVersion - The current version
- * @param {string} options.repoUrl - The GitHub repository URL for the current
- *   project.
- * @param {boolean} options.isReleaseCandidate - Denotes whether the current
- *   project is in the midst of release preparation or not. If this is set, this
- *   command will also ensure the current version is represented in the
- *   changelog with a release header, and that there are no unreleased changes
- *   present.
- * @throws {InvalidChangelogError} Will throw if the changelog is invalid
```
Gudahtt, line 65: Despite the lack of VSCode support, I do think we should preserve these as documentation. They're pretty important for knowing how to properly use the function. As CLI.js does for example.

src/updateChangelog.ts
rekmarks: Done in f2b8390

src/validateChangelog.js
rekmarks: Restored in 78402b7

src/parseChangelog.ts
```diff
@@ -95,10 +108,13 @@ function parseChangelog({ changelogContent, repoUrl }) {
         throw new Error(`Malformed category header: '${truncated(line)}'`);
       }
       const isFirstCategory = mostRecentCategory === null;
-      finalizePreviousChange({ removeTrailingNewline: !isFirstCategory });
-      mostRecentCategory = results[1];
+      finalizePreviousChange({
+        removeTrailingNewline: !isFirstCategory,
+      });
+      // TODO: Get rid of typecast?
+      mostRecentCategory = results[1] as ChangeCategory;
```
rekmarks, line 115: Curious about your thoughts here. I don't think we can be certain that this will be a valid change category?

Gudahtt (dismissed):
LGTM!

src/parseChangelog.ts
```diff
@@ -35,24 +39,31 @@ function parseChangelog({ changelogContent, repoUrl }) {
...
-  function finalizePreviousChange({ removeTrailingNewline = false } = {}) {
+  function finalizePreviousChange({
+    removeTrailingNewline = false,
+  }: {
+    removeTrailingNewline?: boolean;
+  } = {}) {
     if (!currentChangeEntry) {
       return;
     }
+    /* istanbul ignore next: This doesn't happen in practice. */
```
Gudahtt, line 63: I think this could happen with two successive, empty change categories/releases? :thinking:

Gudahtt (approved):
LGTM!

PR #62, opened 2021-05-19
Validate change category during parsing

Gudahtt:
The change category is now validated as part of `parseChangelog`. Previously it was still being validated, but not until later on when the change categories were added to a `Changelog` instance. Validating it earlier in `parseChangelog` makes more sense now that we have types.

### DIFF
--- a/src/parseChangelog.test.ts
+++ b/src/parseChangelog.test.ts
@@ -542,7 +542,7 @@ describe('parseChangelog', () => {
     ).toThrow(`Unrecognized line: '#### Changed'`);
   });
 
-  it('should throw if a change category is unrecognized', () => {
+  it('should throw if a change category is malformed', () => {
     const brokenChangelog = outdent`
       # Changelog
       All notable changes to this project will be documented in this file.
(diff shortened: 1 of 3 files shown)


ryanml (approved):
++

src/parseChangelog.test.ts
```diff
@@ -568,6 +568,32 @@ describe('parseChangelog', () => {
     ).toThrow(`Malformed category header: '### Ch-Ch-Ch-Ch-Changes'`);
```
rekmarks, line 568: > Ch-Ch-Ch-Ch-Changes

😂 💯 👨‍🎤

src/parseChangelog.ts
```diff
@@ -1,10 +1,18 @@
...
+  orderedChangeCategories,
+  unreleased,
+} from './constants';
 
 function truncated(line: string) {
   return line.length > 80 ? `${line.slice(0, 80)}...` : line;
 }
 
+function isValidChangeCategory(category: string): category is ChangeCategory {
+  return orderedChangeCategories.includes(category as ChangeCategory);
```
rekmarks, line 13: I think we could just index into `ChangeCategory` since the values and keys are identical, but hardly worth blocking on.

rekmarks (dismissed):
LGTM! Just one nit you can take or leave.

src/parseChangelog.ts
Gudahtt: Great idea - done in https://github.com/MetaMask/auto-changelog/pull/62/commits/d903e1def68ee4bbb4a9af8d33e2c03920a71059

rekmarks (approved):
🚀

PR #66, opened 2021-05-19
Move PR to end of change entry

Gudahtt:
The PR associated with a change has been moved to the end of the change entry, rather than being used as the prefix. This matches the "Keep A Changelog" style better, and generally makes the changelog easier to read. It also makes it easier to handle the case where one change was related to multiple PRs.


rekmarks (approved):
Excellent!

PR #67, opened 2021-05-19
Update cli-related package.json scripts, fields

rekmarks:
Executing the CLI directly by invoking `./dist/cli.js` wasn't working because we never called `chmod +x` on it. This PR adds a call to `chmod` to the `build` package script.

In addition, it updates the package.json `bin` field to register two names for the application, `@metamask/auto-changelog` and `auto-changelog`. Previously, only the former (i.e. the name of the package) would be usable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,10 @@
     "access": "public"
   },
   "main": "dist/index.js",
-  "bin": "dist/cli.js",
+  "bin": {
+    "@metamask/auto-changelog": "dist/cli.js",
+    "auto-changelog": "dist/cli.js"
+  },
   "engines": {
     "node": ">=12.0.0"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -26,8 +26,8 @@
     "lint:misc": "prettier '**/*.json' '**/*.md' '!CHANGELOG.md' '**/*.yml' --ignore-path .gitignore",
     "lint": "yarn lint:eslint && yarn lint:misc --check",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write",
-    "build": "tsc --project .",
-    "changelog": "node ./src/cli.js"
+    "build": "tsc --project . && chmod +x 'dist/cli.js'",
```
Gudahtt, line 29: Could we move this to the `prepublishOnly` step instead? It's not really necessary during the build. Plus, this command does not work across all platforms. If we're to include non-cross-platform scripts, better that they be kept to the publishing flow if possible.

package.json
```diff
@@ -21,13 +21,13 @@
   "scripts": {
     "test": "jest",
     "test:watch": "jest --watch",
-    "prepublishOnly": "yarn lint && yarn test",
+    "prepublishOnly": "yarn build && chmod +x dist/cli.js && yarn lint && yarn test",
```
Gudahtt, line 24: ... `build` wasn't already here!? :grimacing: Whoops! Missed that too. That could have been unfortunate.

Gudahtt (approved):
LGTM!

PR #69, opened 2021-05-20
Validate changelog in prepublishOnly

rekmarks:
We should probably use the thing in the thing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "scripts": {
     "test": "jest",
     "test:watch": "jest --watch",
-    "prepublishOnly": "yarn build && chmod +x dist/cli.js && yarn lint && yarn test",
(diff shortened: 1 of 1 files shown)


Gudahtt (changes requested):
I'm not sure this is a good idea. The `--rc` flag would mean this would _fail_ after the tag was created, as the tag wouldn't exist in the RC phase. Whereas this command might be run before or after the tag is created (mostly likely after).

Definitely worth adding to CI! But maybe not to `prepublishOnly`

rekmarks:
Guess we'll save it for when we're only publishing in CI 😎

Gudahtt:
Oh I meant on normal CI. Like, alongside unit tests and the linter.

Gudahtt:
e.g. https://github.com/MetaMask/metamask-extension/blob/develop/.circleci/config.yml#L289

PR #76, opened 2021-05-24
Fix changelog where first release isn't first

### LINKED ISSUE #57: Invalid reference link definition when new release is first numerically
When the release being prepared is the _first_ release numerically, the reference link definition generated will be invalid.

For example, if preparing a v0.1.0 release for a project that already has a v1.0.0 release, it will generate this reference link definition: `${repoUrl}/compare/vundefined...v0.1.0`.

In these cases, where we don't have any release to compare to, we just just link the tag itself.

Gudahtt:
`auto-changelog` will now correctly update and validate changelogs where the first release chronologically is different than the first release numerically.

A test has been added to ensure such a changelog is now validated correctly. This test fails on `main`.

Fixes #57

### DIFF
--- a/src/changelog.ts
+++ b/src/changelog.ts
@@ -109,13 +109,16 @@ function stringifyLinkReferenceDefinitions(
   repoUrl: string,
   releases: ReleaseMetadata[],
 ) {
-  const releasesOrderedByVersion = releases
+  // A list of release versions in descending SemVer order
+  const descendingSemverVersions = releases
     .map(({ version }) => version)
     .sort((a: Version, b: Version) => {
       return semver.gt(a, b) ? -1 : 1;
     });
-  const orderedReleases = releases.map(({ version }) => version);
-  const hasReleases = orderedReleases.length > 0;
+  const latestSemverVersion = descendingSemverVersions[0];
+  // A list of release versions in chronological order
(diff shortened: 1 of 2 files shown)


src/changelog.ts
```diff
@@ -139,20 +139,21 @@ function stringifyLinkReferenceDefinitions(
   // patch releases on older releases can be accomodated.
   const releaseLinkReferenceDefinitions = releases
     .map(({ version }) => {
+      let diffUrl;
       if (version === orderedReleases[orderedReleases.length - 1]) {
```
rekmarks, line 143: My only feedback is that `orderedReleases` (and really the releases param) could be documented / named such that it's immediately obvious that the releases are sorted in chronological order.k

rekmarks (approved):
LGTM! Just one nit.

src/changelog.ts
Gudahtt: Good call, I'll rename this

Gudahtt: Done in https://github.com/MetaMask/auto-changelog/pull/76/commits/027e394c92bc254503aa153ad009ea6541f57282

rekmarks (approved):
I'm sold.

PR #81, opened 2021-05-25
Use `actions/setup-node@v2` instead of `actions/setup-node@v1`

Gudahtt:
The latest version of `actions/setup-node@v2` appears to act the same, except that it has better caching ([see here](https://github.com/actions/setup-node#v2)).


rekmarks (approved):
LGTM!

PR #83, opened 2021-05-25
Fix PR number detection in existing change entries

Gudahtt:
The `update` command will filter out any commits that are the result of PRs that have been referenced already in the changelog.

Unfortunately this PR number detection only looked at the beginning of the entry. This was because prior to #66, that's where it would be by default (though this was always a poor decision, because it was always possible to update the change entry and reference the PR anywhere). This is especially broken since #66.

We now look for things that look like PR numbers anywhere in the change description.


Gudahtt:
I noticed that we aren't detecting PR numbers correctly if multiple are referenced in one change description - we only find the first one. I'll address that in a later PR.

rekmarks (approved):

PR #84, opened 2021-05-25
Detect all PRs referenced in each change description

Gudahtt:
We now detect all referenced PRs in each change description, rather than just the first. This is especially helpful when one change entry references many PRs (e.g. a set of related bug fixes).

Manual testing steps:
- Checkout this branch and run `yarn build`
- Run `yarn changelog update` a few times and observe how the duplicate detection works in the typical case where only one PR is referenced. After running this a few times the changelog should have one copy of each commit with a PR number, but multiple for the commit on this branch which has no number yet.
- Combine all unreleased change entries that include a PR number into a single change entry (e.g. indent them all once, that should do it).
- Run `yarn changelog update` again a few times and observe that the duplication detection still works the same as before.


Gudahtt:
This depends upon #83

rekmarks (approved):
Nice testing instructions. LGTM!

PR #86, opened 2021-05-26
Add more exports to index.ts

rekmarks:
I found myself in a position where I wanted to use `parseChangelog`. While we're here, we might as well export `createEmptyChangelog` and the `Changelog` class, as well. `import type { Changelog }` could be useful in TypeScript projects, for example.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,6 +1,7 @@
 module.exports = {
   collectCoverage: true,
   collectCoverageFrom: ['./src/**.ts'],
+  coveragePathIgnorePatterns: ['/node_modules/', 'index.ts'],
   coverageReporters: ['text', 'html'],
   coverageThreshold: {
     global: {
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,7 +1,9 @@
 module.exports = {
   collectCoverage: true,
   collectCoverageFrom: ['./src/**.ts'],
-  coveragePathIgnorePatterns: ['/node_modules/', 'index.ts'],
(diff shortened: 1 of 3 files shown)


Gudahtt (approved):
LGTM!

I was hesitant to expose this stuff at first because we didn't have a use for it, and more exports means more changes to break things. But, I agree, these seem useful.

jest.config.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   collectCoverage: true,
   collectCoverageFrom: ['./src/**.ts'],
+  coveragePathIgnorePatterns: ['/node_modules/', 'index.ts'],
```
rekmarks, line 4: The `jest` coverage reporter (i.e. `istanbul`) registers every exported function in `index.ts` as a distinct function, tanking our function coverage.

Gudahtt (approved):

jest.config.js
```diff
@@ -1,6 +1,9 @@
 module.exports = {
   collectCoverage: true,
   collectCoverageFrom: ['./src/**.ts'],
+  // Jest (i.e. istanbul) identifies every export in index.ts as a distinct
+  // function, which tanks our function coverage.
+  coveragePathIgnorePatterns: ['index.ts'],
```
Gudahtt, line 6: Kinda think we should still count this in our coverage report though :thinking: This is what e2e tests would cover. It's important to make sure our exports exist and are named correctly and so on. This would catch the ES interop issue we've had many times as well, where we broke exports accidentally.

This is fine for now though.

PR #87, opened 2021-05-28
Remove projectRootDirectory param from getMostRecentTag

rekmarks:
The `projectRootDirectory` param was added to a call to `git rev-list` in `getMostRecentTag`. Per the [`git rev-list` docs](https://git-scm.com/docs/git-rev-list), providing one or more paths to that command will limit the returned lists to commits "modifying the given <paths>". This param is not only unnecessary, but prevents us from retrieving tags belonging to merge commits, because they don't modify anything. The purpose of `getMostRecentTag` is to get the most recent tag for the entire repository, not for a particular sub-tree.


Gudahtt (approved):
LGTM!

PR #92, opened 2021-06-23
Permit missing separating space in unreleased link reference definition

rekmarks:
This enables the parsing of changelogs without a separating space in their `Unreleased` link reference definition. Changelog updating and validation remain unaffected by this change.

This package is very strict about formatting, however, we can afford to be somewhat lenient during parsing, and I think it's worthwhile to spare users from manually adding spaces to their link reference definitions.


Gudahtt (approved):
LGTM!

PR #94, opened 2021-06-28
Replace runCommand with execa

rekmarks:
Replaces the local `runCommand` function a tiny wrapper around `execa`. Adds an additional call to `assert` to exit early in case of a git failure.


Gudahtt (approved):
LGTM!

PR #102, opened 2021-06-30
Validate no uncategorized changes in RC

Gudahtt:
The validation command now ensures there are no uncategorized changes in RCs.

A few extra tests have been added to prevent coverage from dropping, due to how the code was reorganized.


brad-decker (approved):
Test cases make it very clear what is going on here. Well done LGTM

Issue #90, opened 2021-06-03
Add `--fix` option to `validate` command


Gudahtt:
The `validate` command doesn't currently allow fixing any issues it finds, even in cases where it would be relatively easy. The workaround for the moment is to use `update` to fix formatting issues (or fix them manually), but `update` also updates the changelog which might be undesirable.

We should add a `--fix` flag to perform apply any fixes we can, in the same manner that `update` does except without the updating.

PR #106, opened 2021-07-13
Skip validation of "Uncategorized" header for older releases

Gudahtt:
Currently the `validate` command will ensure that no release has uncategorized changes when the `--rc` flag is present, but it doesn't care about uncategorized changes when the `--rc` flag isn't present.

This is against the spirit of the `--rc` flag. That flag should enable extra validation specific to the preparation of the current release. It shouldn't impose further burdens on maintainers that have nothing to do with the current release. This validation is also very inconvenient for some projects (e.g. `metamask-extension`), where the current maintainers may not have all of the context to make improvements to ancient changelog entries.

### DIFF
--- a/src/validate-changelog.test.ts
+++ b/src/validate-changelog.test.ts
@@ -618,7 +618,7 @@ describe('validateChangelog', () => {
       ).toThrow('Uncategorized changes present in the changelog');
     });
 
-    it('should throw if there are uncategorized changes in an older release', () => {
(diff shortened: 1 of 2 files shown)


shanejonas:
would be nice to have an option to force the validation on new repos

shanejonas (approved):

Gudahtt:
That would be useful! There is an issue open for validation: https://github.com/MetaMask/auto-changelog/issues/90

PR #111, opened 2022-05-16
Read repository from package.json if `npm_package_repository_url` is not set

Mrtenz:
In Yarn 3, `npm_package_repository_url` isn't set, and `auto-changelog` fails to parse the repository. In this PR I added a function which first looks at the `npm_package_repository_url`, and if that doesn't exist, looks at `PROJECT_CWD` instead. From there the `package.json` can be parsed in order to get the `repository` automatically.

For details on Yarn 3 environment variables, see [here](https://yarnpkg.com/advanced/lifecycle-scripts#environment-variables).

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -23,6 +23,7 @@ import {
   InvalidChangelogError,
   validateChangelog,
 } from './validate-changelog';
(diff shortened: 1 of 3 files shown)


rekmarks (approved):
LGTM!

PR #113, opened 2022-06-01
Account for tiny time differences between commits

mcmire:
When two tags are created in a repo that differ by mere milliseconds or
smaller, such as would happen in an automated high-level test of the
`create-release-pr` action, `git rev-list --tags` seems to get confused,
and so `getMostRecentTag` may not always return the latest tag. Adding
`--date-order` seems to fix this.


rekmarks (approved):
LGTM!

PR #117, opened 2022-10-28
Update minimum Node.js version to v14

Gudahtt:
The minimum Node.js version has been updated from v12 to v14, and CI has been updated to test v18.


.github/workflows/build-lint-test.yml
```diff
@@ -39,10 +39,12 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js v12
-        uses: actions/setup-node@v2
+      - name: Get Node.js version
+        id: nvm
+        run: echo ::set-output name=NODE_VERSION::$(cat .nvmrc)
```
Mrtenz, line 44: Can you use the non-deprecated way of doing this?

.github/workflows/build-lint-test.yml
```diff
@@ -39,10 +39,12 @@ jobs:
...
+      - uses: actions/setup-node@v2
```
Mrtenz, line 45: Maybe bump all actions to v3 as well while you're at it.

Mrtenz (changes requested):

Mrtenz (approved):

PR #120, opened 2022-10-31
Support alternate tag prefixes

### LINKED ISSUE #116: Support package-level tags
Currently when looking for tags, we assume that all tags start with `v`: https://github.com/MetaMask/auto-changelog/blob/1c931ec77867f2d70ff5197cc25ee7dba99d2425/src/update-changelog.ts#L19

We want this to be configurable. Perhaps we accept a `tagPrefix` option that's an array of prefixes to try.

Gudahtt:
We should support alternate prefixes in version tags. At the moment the project assumes that the prefix `v` is always used. But some projects use no prefix, or include the package as a prefix to the version (e.g. in monorepos).

Fixes #116


Gudahtt:
I will follow up with manual testing steps soon, and do some manual testing myself.

Gudahtt:
Manual testing steps:

## Independent repository (regression testing):

* Checkout `auto-changelog` and build the project
* Add the script `"auto": "[path-to-auto-changelog-project]/dist/cli.js" to some repository for test purposes
  * It can be any of our projects, as long as it's not a monorepo, and as long as it has some changes and some previous git tag.
* Run `yarn auto update --rc`
* See that it correctly identified the changes since the last tag, and it correctly set the release header in the changelog.

## Monorepo repository (new functionality):

* Checkout `auto-changelog` and build the project
* Checkout the controllers repository `monorepo-poc` branch, and build the project
* Add the script `"auto": "[path-to-auto-changelog-project]/dist/cli.js" to the root of the repository.
* Run `yarn auto update --rc --root ./packages/base-controller --tag-prefix=base-controller@v`
* See that it correctly identified the changes since the last tag, and it correctly set the release header in the changelog.
* Commit the changelog changes and tag that commit as `base-prefix@v[version]`
* Make some example commits that alter the base controller package
* Run `yarn auto update --rc --root ./packages/base-controller --tag-prefix=base-controller@v`
* See that it correctly identifies just the example commits

Gudahtt:
The manual testing steps include setting the `--currentVersion` flag because the default version used is `npm_project_version`, which is wrong if the update script is being run from the root of the monorepo.

This seems wrong to me (the version should always match the manifest at the project root), so I have updated how that works here: https://github.com/MetaMask/auto-changelog/pull/121

rickycodes (approved):
This all LGTM

PR #121, opened 2022-11-02
Get package version from manifest

Gudahtt:
The package version is now obtained from the package manifest rather than from the `npm_package_version` environment variable. This strategy works better with the `--root` flag, which might sometimes specify a root project directory that is different from the current working directory. The current version should always match the version in the manifest at the project root.


src/cli.ts
```diff
@@ -327,6 +304,55 @@ async function main() {
...
+        exitWithError(
+          `Access to package manifest is forbidden by file access permissions: '${manifestPath}'`,
+        );
+        return;
+      } else if (error.name === 'SyntaxError') {
+        exitWithError(
+          `Package manifest cannot be parsed as JSON: '${manifestPath}'`,
+        );
+        return;
+      }
```
legobeat, line 334:
```suggestion
      if (error.code === 'ENOENT') {
        return exitWithError(
          `Package manifest not found at path: '${manifestPath}'\nRun this script from the project root directory, or set the project directory using the '--root' flag.`,
        );
      }
      if (error.code === 'EACCES') {
        return exitWithError(
          `Access to package manifest is forbidden by file access permissions: '${manifestPath}'`,
        );
      }
      if (error.name === 'SyntaxError') {
        return exitWithError(
          `Package manifest cannot be parsed as JSON: '${manifestPath}'`,
        );
      }
      throw err;
```

src/cli.ts
```diff
@@ -327,6 +304,55 @@ async function main() {
...
+    return;
+  } else if (!repoUrl) {
+    exitWithError(
+      `npm package repository URL not found. Please set the '--repo' flag, or run this as an npm script from a project with the 'repository' field set.`,
+    );
+    return;
+  } else if (!isValidUrl(repoUrl)) {
+    exitWithError(`Invalid repo URL: '${repoUrl}'`);
+    return;
+  }
```
legobeat, line 354:
```suggestion
    return exitWithError(
      `Version not found. Please set the --currentVersion flag, or run this as an npm script from a project with the 'version' field set.`,
    );
  }
  if (semver.valid(currentVersion) === null) {
    return exitWithError(`Current version is not valid SemVer: '${currentVersion}'`);
    return;
  }
  if (!repoUrl) {
    exitWithError(
      `npm package repository URL not found. Please set the '--repo' flag, or run this as an npm script from a project with the 'repository' field set.`,
    );
    return;
  }
  if (!isValidUrl(repoUrl)) {
    return exitWithError(`Invalid repo URL: '${repoUrl}'`);
  }
```

legobeat (changes requested):
Note in particular rethrowing errors that aren't explicitly caught.

src/cli.ts
Gudahtt: The pattern of placing the return on the following line was to make complying with the ESLint "consistent-return" rule easier.

Now that you're pointing it out, I think I agree that this style is harder to read and more error prone. I can adjust that in a follow-up PR and find a better solution for complying with that lint rule. I'd rather avoid doing it here because the same pattern is used in more places, and it's distinct from the main purpose of this PR.

Gudahtt:
Oh darn, great find! Not re-throwing the error is a bad one. I've just fixed that in https://github.com/MetaMask/auto-changelog/pull/121/commits/41cb1e6fa883eb9a002eea745d504cb5f7dd61e4

The requested style changes I can address in a follow-up PR

src/cli.ts
Gudahtt: Done in #122

legobeat (approved):
lgtm!

PR #122, opened 2022-11-04
Refactor: `return` on same line as error

Gudahtt:
The CLI has been refactored to `return` on the same line an error is handled (by logging and setting the exit code). This is easier to read and less error prone (reduced risk of forgetting to `return`).


Gudahtt:
This depends upon #121

legobeat (approved):

PR #124, opened 2022-11-04
Update TypeScript to the latest stable version

Gudahtt:
TypeScript has been updated to the latest stable version. Additionally, the version range has been updated to use `~` because TypeScript does not follow SemVer (they make breaking changes as minor updates).

The configuration has been updated to match the module template, except that two options (`exactOptionalPropertyTypes` and `noUncheckedIndexedAccess`) have been omitted temporarily to reduce the number of changes required for this update. They can be added in later PRs.

Just as in the module template, a separate `tsconfig` file has been created to distinguish build settings from settings used to compile unit tests.

The one configuration difference is the compilation target. The target is ES2019 here because this is a developer tool, so we don't need to stick to ES2017 for browser compatibility.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -183,6 +183,19 @@ async function validate({
   }
 }
 
+/**
+ * Returns whether an error has an error code or not.
+ *
+ * @param error - The error to check.
+ * @returns True if the error is a real error and has a code property, false otherwise.
+ */
+function hasErrorCode(error: unknown): error is Error & { code: unknown } {
+  return (
+    error instanceof Error &&
+    Object.prototype.hasOwnProperty.call(error, 'code')
+  );
+}
+
 type InitOptions = {
   changelogPath: string;
   repoUrl: string;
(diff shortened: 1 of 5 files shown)


Gudahtt:
Issue created here to track the additional TypeScript options: https://github.com/MetaMask/auto-changelog/issues/127

src/cli.ts
```diff
@@ -328,18 +343,20 @@ async function main() {
...
+          );
+        } else if (error.code === 'EACCES') {
+          return exitWithError(
+            `Access to package manifest is forbidden by file access permissions: '${manifestPath}'`,
+          );
+        } else if (error.name === 'SyntaxError') {
+          return exitWithError(
+            `Package manifest cannot be parsed as JSON: '${manifestPath}'`,
+          );
+        }
```
Mrtenz, line 359: Shouldn't this be moved to outside the `hasErrorCode` block?

Gudahtt: Ah, good catch, I will fix this

Gudahtt: Done in d16296b63f1da4bb098f834230f52821dc2ad5a0

Mrtenz (approved):

PR #130, opened 2022-11-15
Allow prerelease versions in release headers

mcmire:
Currently, this tool expects the version in a release header to be a version that looks like `<major>.<minor>.<patch>` and rejects other formats. However, it may be useful to publish a prerelease version of a package so that integration with other products/services can be tested in a more "live" manner, particularly if that package is a part of a monorepo. A prerelease version looks like
`<major>.<minor>.<patch>-<prerelease identifier>.<number>`, where `<prerelease identifier>` can be an arbitrary word like `alpha`, `beta`, or `rc`. This commit allows such versions.

### DIFF
--- a/src/parse-changelog.test.ts
+++ b/src/parse-changelog.test.ts
@@ -199,6 +199,45 @@ describe('parseChangelog', () => {
     expect(changelog.getUnreleasedChanges()).toStrictEqual({});
   });
 
+  it('should parse changelog with prereleases', () => {
+    const changelog = parseChangelog({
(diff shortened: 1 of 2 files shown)


mcmire:
This commit should hopefully fix the CI errors we're seeing here: https://github.com/MetaMask/controllers/actions/runs/3473161747/jobs/5804878754#step:6:14

FrederikBolding (approved):

Gudahtt (approved):
LGTM!

PR #146, opened 2023-05-09
deps: execa@^5.1.1->^7.1.1

legobeat:


mcmire:
@legobeat This might be challenging. We have intentionally been using an older version of `execa` across our repos because newer versions only support ESM, and our projects expect CommonJS: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

PR #148, opened 2023-06-23
Add `--fix` option to validate command

### LINKED ISSUE #90: Add `--fix` option to `validate` command

Mrtenz:
This adds a `--fix` option to the `auto-changelog validate` command, which writes a new changelog file if it can be fixed. If `--fix` is not specified, the diff is shown as usual.

Closes #90.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -134,6 +134,7 @@ type ValidateOptions = {
   isReleaseCandidate: boolean;
   repoUrl: string;
   tagPrefix: string;
+  fix: boolean;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


FrederikBolding (approved):

Gudahtt (approved):
LGTM!

FrederikBolding:
> Wow, you know I suspected this would be easy but never looked into it. I wish I had! This would have saved a lot of time.

Agreed. I have manually fixed the changelog formatting too many times 😅

Issue #60, opened 2021-05-19
Incompatibilities with some markdownlint rules


rekmarks:
This package produces `CHANGELOG.md` files that are incompatible with [`markdownlint`](https://github.com/DavidAnson/markdownlint).

The written changelog conforms exactly to the example from the [Keep a Changelog website](https://keepachangelog.com/en/1.0.0/), but are its spacing conventions really relevant to the spec our our ability to parse a changelog?

I believe that the spacing is irrelevant to the Keep a Changelog spec, and we should make the output changelogs compatible with `markdownlint` at least in terms of spacing. There may even be an argument to integrate `markdownlint` into this tool, but that's a different question.

Related: #50. `markdownlint`-compatibility appears to be largely equivalent with Prettier-compatibility.

Gudahtt:
Changing the spacing, or some other unimportant aspect of the changelog format, is something we could do. But then it wouldn't be strictly compatible with the spec, and thus with other KeepAChangelog-related tooling. We could make it flexible, but then it would be less useful, because it can't apply the format for us automatically if it's flexible enough to accept both.

danroc:
Update: Keep a Changelog has changed its own [changelog file](https://raw.githubusercontent.com/olivierlacan/keep-a-changelog/v1.1.1/CHANGELOG.md) to match `markdownlint` defaults.

See: https://github.com/olivierlacan/keep-a-changelog/pull/318

PR #155, opened 2023-09-22
Add `--prettier` flag for Prettier-formatted changelogs

danroc:
This PR adds a new `--prettier` flag. When enabled, this flag signals that the tool should expect the changelog to be formatted using Prettier. By default, this flag is set to `false`, ensuring backward compatibility with all existing `CHANGELOG.md` files.

The primary motivation for this change is to overcome a current limitation of `auto-changelog`. At present, our tool does not support the changelog format provided by _Keep a Changelog_ (version ^1.1.0).

Fixes: #60

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "eslint-plugin-prettier": "^4.2.1",
     "jest": "^26.4.2",
     "outdent": "^0.8.0",
-    "prettier": "^3.0.3",
+    "prettier": "^2.7.1",
     "rimraf": "^3.0.2",
     "ts-jest": "^26.5.6",
     "typescript": "~4.8.4"
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "eslint-plugin-prettier": "^4.2.1",
     "jest": "^26.4.2",
     "outdent": "^0.8.0",
-    "prettier": "^2.8.8",
+    "prettier": "^3.0.3",
     "rimraf": "^3.0.2",
     "ts-jest": "^26.5.6",
     "typescript": "~4.8.4"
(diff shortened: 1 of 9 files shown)


Mrtenz (commented):
Rather than running Prettier programatically, couldn't we just remove the changelog from the ignore list, and disable formatting with a flag?

https://github.com/MetaMask/metamask-module-template/blob/main/package.json#L45

danroc:
I'm not sure. The `update` command would need to be paired with an external formater everytime it's executed. But it could be another flag (`--ignore-format`?) available only for `validate`.

Mrtenz:
> I'm not sure. The `update` command would need to be paired with an external formater everytime it's executed. But we could add another flag (`--ignore-formatting`?) available only for `validate`.

We run Prettier regardless, hence why it's currently ignored.

Gudahtt:
Not always at the same time as updating the changelog though.

There's also the `validate` command; it's sensitive to formatting changes so we'd need to run prettier programmatically during validation to avoid breaking that.

danroc:
If we ignore formating, enforcing that the sections are always in the same order may be more complex.

Yes @Gudahtt, this is why I passed the formatter to both `validate` and `update` commands.

Mrtenz:
Ok, makes sense!

danroc:
If we are not willing to add a formatter, I suppose I could workaround like this:

Replace `update` with:

```
auto-changelog update && prettier
```

And replace `validate` with:

```
auto-changelog validate --fix && prettier
```

and then check if the `CHANGELOG.md` file changed.

Gudahtt (approved):
LGTM!

PR #157, opened 2023-10-06
package rename changelog validation

kanthesha:
In the current released version, if we rename the package and would like to preserve the changelog, validation is failing for the format. Which is expected as package name doesn't match.

These changes will help to preserve the old package release/tag history and also works fine with new package release/s when we run the validation.
It addresses
1. Changelog validation working fine for the old package name in the entries and also for the new package release entries
2. An option to configure Old Tag Prefix and Old Package Version

original tag prefix and original latest version can be configured in package.json
```
"autoChangelog": {
    "packageRename": {
      "originalLastestVersion": "5.0.1",
      "originalTagPrefix": "json-rpc-middleware-stream@"
    }
  },
```

### DIFF
--- a/src/changelog.ts
+++ b/src/changelog.ts
@@ -271,7 +271,7 @@ function getReleaseLinkReferenceDefinitions(
   const releaseLinkReferenceDefinitions = releases.map(({ version }) => {
     let diffUrl;
     // once the version matches with versionBeforeRename, rest of the lines in changelog will be assumed as migrated tags
-    if (packageRename && packageRename?.versionBeforeRename === version) {
+    if (packageRename && packageRename.versionBeforeRename === version) {
       tagPrefixToCompare = packageRename.tagPrefixBeforeRename;
     }
 


src/changelog-config.ts
```diff
@@ -0,0 +1,63 @@
...
+      originalTagPrefix: string;
+    };
+  };
+};
+
+/**
+ * Returns the latest version of the original package in case of package renamed.
+ *
+ * @returns The original package latest version.
+ */
```
MajorLift, line 19: `@returns` tag should include return types for all branches.
```suggestion
 * @returns The latest version string of the original package, or null.
```

src/changelog-config.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (
+      typeof packageJsonContent.autoChangelog?.packageRename
+        ?.originalTagPrefix === 'string'
+    ) {
+      return packageJsonContent.autoChangelog.packageRename.originalTagPrefix;
+    }
+  }
+
+  return null;
+}
```
MajorLift, line 63: A couple changes here might reduce the complexity of the logic and add a bit of type safety:
- Early exit in case of `undefined` `packageJsonPath` -> avoids unnecessary nesting.
- Replacing `as` type assertion for with type annotation.
- By removing the `as` assertion and typing `packageJsonContent` as `PackageJson`, we no longer need to check whether `originalTagPrefix` is not a string, since if that happens it will trigger a "not assignable" type error on line 54.

```suggestion
  if (packageJsonPath === undefined) {
    return null;
  }

  const packageJson = path.resolve(packageJsonPath);
  const packageJsonContent: PackageJson = JSON.parse(
    fs.readFileSync(packageJson, 'utf8'),
  );

  return (
    packageJsonContent.autoChangelog?.packageRename?.originalTagPrefix ?? null
  );
```

src/changelog.ts
```diff
@@ -192,35 +195,71 @@ function stringifyLinkReferenceDefinitions(
...
+  let tagPrefixToCompare = tagPrefix;
+  // if there is a package renamed and version and tag prefix set in package.json
+  // original package tag prefix will be considered for compare
+  // below if is for an example from changelog [Unreleased]: https://github.com/MetaMask/core/compare/json-rpc-middleware-stream@5.0.1...HEAD
+  if (
+    orgTagPrefix &&
+    orgLatestVersion &&
+    orgLatestVersion === latestSemverVersion
+  ) {
+    tagPrefixToCompare = orgTagPrefix;
```
MajorLift, line 209: The second null check is redundant, since it's covered by the the deep equals comparison between `null` and `undefined`.
```suggestion
  if (orgTagPrefix && orgLatestVersion === latestSemverVersion) {
```

src/changelog.ts
```diff
@@ -192,35 +195,71 @@ function stringifyLinkReferenceDefinitions(
...
-          : getTagUrl(repoUrl, `${tagPrefix}${version}`);
+        // if there is a package renamed and version and tag prefix set in config
+        // this if condition will fix the validation for original package's first release from new/renamed package
+        // [6.0.0]: https://github.com/MetaMask/core/compare/json-rpc-middleware-stream@5.0.1...@metamask/json-rpc-middleware-stream@6.0.0
+        if (orgLatestVersion && orgLatestVersion === previousVersion) {
+          diffUrl = previousVersion
+            ? getCompareUrl(
+                repoUrl,
+                `${orgTagPrefix as string}${previousVersion}`,
+                `${tagPrefixToCompare}${version}`,
```
MajorLift, line 251: - The `orgLatestVersion` null check here is also redundant.
- We can remove the `as` assertion here as well by doing a null check for `orgTagPrefix` first.
```suggestion
        if (orgTagPrefix && orgLatestVersion === previousVersion) {
          diffUrl = previousVersion
            ? getCompareUrl(
                repoUrl,
                `${orgTagPrefix}${previousVersion}`,
```

src/changelog-config.ts
```diff
@@ -0,0 +1,63 @@
...
+      originalLastestVersion: string;
+      originalTagPrefix: string;
+    };
+  };
+};
+
+/**
+ * Returns the latest version of the original package in case of package renamed.
+ *
+ * @returns The original package latest version.
```
MajorLift, line 18: `@returns` tag should include return types for all branches.
```suggestion
 * @returns The latest version string of the original package, or null.
```

src/changelog-config.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  return null;
+}
+
+/**
+ * Returns the original package tag prefix in case of package renamed.
+ *
+ * @returns The original package tag prefix.
```
MajorLift, line 44: `@returns` tag should include return types for all branches.
```suggestion
 * @returns The tag prefix string for the original package, or null.
```

src/changelog.ts
```diff
@@ -192,35 +195,71 @@ function stringifyLinkReferenceDefinitions(
...
+  const orgTagPrefix = getOriginalTagPrefix();
+  let tagPrefixToCompare = tagPrefix;
+  // if there is a package renamed and version and tag prefix set in package.json
+  // original package tag prefix will be considered for compare
+  // below if is for an example from changelog [Unreleased]: https://github.com/MetaMask/core/compare/json-rpc-middleware-stream@5.0.1...HEAD
+  if (
+    orgTagPrefix &&
+    orgLatestVersion &&
+    orgLatestVersion === latestSemverVersion
+  ) {
```
MajorLift, line 208: The second null check is redundant, since it's covered by the the deep equals comparison between `null` and `undefined`.
```suggestion
  if (orgTagPrefix && orgLatestVersion === latestSemverVersion) {
```

src/changelog.ts
```diff
@@ -192,35 +195,71 @@ function stringifyLinkReferenceDefinitions(
...
-            )
-          : getTagUrl(repoUrl, `${tagPrefix}${version}`);
+        // if there is a package renamed and version and tag prefix set in config
+        // this if condition will fix the validation for original package's first release from new/renamed package
+        // [6.0.0]: https://github.com/MetaMask/core/compare/json-rpc-middleware-stream@5.0.1...@metamask/json-rpc-middleware-stream@6.0.0
+        if (orgLatestVersion && orgLatestVersion === previousVersion) {
+          diffUrl = previousVersion
+            ? getCompareUrl(
+                repoUrl,
+                `${orgTagPrefix as string}${previousVersion}`,
```
MajorLift, line 250: - The `orgLatestVersion` null check here is also redundant.
- We can remove the `as` assertion here as well by doing a null check for `orgTagPrefix` first.
```suggestion
        if (orgTagPrefix && orgLatestVersion === previousVersion) {
          diffUrl = previousVersion
            ? getCompareUrl(
                repoUrl,
                `${orgTagPrefix}${previousVersion}`,
```

MajorLift:
Looks good overall, especially with how quickly you got this to work!
Maybe you could add a simple test case in `changelog.test.ts` to ensure that `stringifyLinkReferenceDefinitions` doesn't break in the future?

src/changelog-config.ts
```diff
@@ -0,0 +1,63 @@
...
+    ) as PackageJson;
+    if (
+      typeof packageJsonContent.autoChangelog?.packageRename
+        ?.originalTagPrefix === 'string'
+    ) {
+      return packageJsonContent.autoChangelog.packageRename.originalTagPrefix;
+    }
+  }
+
+  return null;
```
MajorLift, line 62: A couple changes here could simplify the logic and add type safety:

```suggestion
  if (packageJsonPath === undefined) {
    return null;
  }

  const packageJson = path.resolve(packageJsonPath);
  const packageJsonContent: PackageJson = JSON.parse(
    fs.readFileSync(packageJson, 'utf8'),
  );

  const { originalTagPrefix } = packageJsonContent.autoChangelog?.packageRename ?? {}
  return typeof originalTagPrefix === 'string' ? originalTagPrefix : null
```

- Early exit in case of `undefined` `packageJsonPath` -> less nesting, explicit handling of edge case.
- Replacing the `as` type assertion for `packageJsonContent` with a type annotation. 
- Replacing long optional chaining expression for `typeof` check and return statement with `originalTagPrefix` variable.
  - Safely destructuring `originalTagPrefix` from possible `undefined` using `??` operator.

src/changelog-config.ts
kanthesha: Sure, adding the return types.
```suggestion
 * @returns {string|null} The latest version of the original package, or null.
```

src/changelog-config.ts
kanthesha: Sure, adding the return types.
```suggestion
 * @returns {string|null} The tag prefix for the original package, or null.
```

src/changelog-config.ts
kanthesha: 1. Being a small function, both processing wise and readability, I don't see change can make much difference! but I think this check `if(packageJsonPath)` will better suit here compare to `packageJsonPath === undefined`. The first check will also take care in case of null.
2. Even if it's undefined, the === 'string' will take care. It'll return when the value is set for the property otherwise null

MajorLift: - Future versions of node will enforce a `?: string | number | boolean` type for the `process.env` values, so I don't think a truthy vs. falsy check is safe here. 
- Currently, a `null` value assigned as a `process.env` will be implicitly converted to the string `"null"`, and future versions will throw an error for a `null` value, so a deep equals check for `undefined` seems sufficient.
- We might also want to put line 55 (`fs.readFileSync()`) inside of a try catch block to guard against invalid paths. 
- Destructuring `originalTagPrefix` isn't necessary, but it has the advantage of letting us avoid repeating the long optional chaining expression, especially if there's any chance of more code being added down the road that accesses the `originalTagPrefix` property.

src/changelog-config.ts
MajorLift: There's a `jsdoc/no-types` rule in `@metamask/eslint-config-typescript` that prevents the `{[type]}` syntax from being used in jsdoc `@returns` tags.

kanthesha: As this package doesn't use the `@metamask/eslint-config-typescript`, lint didn't complain. But I can see that in core and it's better to be consistent.

src/changelog.ts
```diff
@@ -192,35 +195,67 @@ function stringifyLinkReferenceDefinitions(
   //
   // If there have not been any releases yet, the repo URL is used directly as
   // the link definition.
+  const orgLatestVersion = getOriginalLatestVersion();
+  const orgTagPrefix = getOriginalTagPrefix();
+  let tagPrefixToCompare = tagPrefix;
```
mcmire, line 200: It looks like we define this variable here and then change it as we loop over versions below. It feels like that could make this code more difficult to follow in the future.

Other than the changes you've made, it looks like the code responsible for defining `unreleasedLinkReferenceDefinition` and `releaseLinkReferenceDefinitions` are both fairly self-contained. What do you think about extracting these to functions? This way you can keep the definition of `tagPrefixToCompare` scoped to each of these functions.

mcmire (commented):
I wonder if using a config file is the right way to go here? I've left some comments below.

src/changelog-config.ts
MajorLift: `auto-changelog` extends `@metamask/eslint-config-typescript` as well. You should be seeing the linter error when you add `{string|null}`. Maybe try restarting the eslint server in vscode?

src/changelog-config.ts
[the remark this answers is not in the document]
kanthesha: First I would like to address, how it works.
I didn't get the clear reason, why it will not work in monorepo! Yeah, monorepo contains multiple packages, so the auto change log config will be added in the each package's package.json. As I sent in the slack message, when I started working on this, there're 3 ways we can pass it, command line is one of them. My thought process was that separate config object will give a cleaner option even when the number of parameters will increase.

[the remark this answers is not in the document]
kanthesha: There's no separate config file for this, the parameters will be within the package's package.json.

src/changelog.ts
kanthesha: When we look at the function `stringifyLinkReferenceDefinitions`, the entire function is only about these two `unreleasedLinkReferenceDefinition` and `releaseLinkReferenceDefinitions`. Yeah, I agree that the function needs a refactor as it's already touching 100 lines mark. I can create a ticket for that.

src/changelog-config.ts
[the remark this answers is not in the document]
kanthesha: And right now we're only trying to implement for the one time package rename! But the ideal feature and implementation would be to work for package renaming even when it is multiple times, which requires an array of {version: string, tagPrefix: string}.

src/changelog.ts
kanthesha: @mcmire the refactor has been done.

src/changelog.ts
```diff
@@ -264,28 +354,42 @@ export default class Changelog {
 
   #formatter: Formatter;
 
+  readonly #versionBeforePkgRename: string | undefined;
+
+  readonly #tagPrefixBeforePkgRename: string | undefined;
+
```
MajorLift, line 360: With `exactOptionalPropertyTypes` disabled, the `| undefined` is redundant.
```suggestion
  readonly #versionBeforePkgRename?: string;

  readonly #tagPrefixBeforePkgRename?: string;
```

src/changelog.ts
```diff
@@ -264,28 +354,42 @@ export default class Changelog {
...
     formatter = (changelog) => changelog,
+    versionBeforePkgRename = undefined,
+    tagPrefixBeforePkgRename = undefined,
   }: {
     repoUrl: string;
     tagPrefix?: string;
     formatter?: Formatter;
+    versionBeforePkgRename?: string | undefined;
+    tagPrefixBeforePkgRename?: string | undefined;
   }) {
```
MajorLift, line 385: With `exactOptionalPropertyTypes` disabled, the `| undefined` is redundant.
```suggestion
    versionBeforePkgRename?: string;
    tagPrefixBeforePkgRename?: string;
```

src/changelog.ts
MajorLift, line 1: Could we add these changes to remove the `as` assertions and make the code a little less brittle against updates in `ChangeCategory`?
```diff
@@ -487,11 +487,13 @@ export default class Changelog {
 
     const unreleasedChanges = this.#changes[unreleased];
 
-    for (const category of Object.keys(unreleasedChanges) as ChangeCategory[]) {
+    for (const category of Object.keys(
+      unreleasedChanges,
+    ) as (keyof typeof unreleasedChanges)[]) {
       if (releaseChanges[category]) {
         releaseChanges[category] = [
-          ...(unreleasedChanges[category] as string[]),
-          ...(releaseChanges[category] as string[]),
+          ...(unreleasedChanges[category] ?? []),
+          ...(releaseChanges[category] ?? []),
         ];
       } else {
         releaseChanges[category] = unreleasedChanges[category];
```

src/changelog.ts
```diff
@@ -264,28 +354,42 @@ export default class Changelog {
 
   #formatter: Formatter;
 
+  readonly #versionBeforePkgRename: string | undefined;
+
+  readonly #tagPrefixBeforePkgRename: string | undefined;
```
MajorLift, line 359: Consistent typing for optional properties. 
```suggestion
  readonly #versionBeforePkgRename?: string;

  readonly #tagPrefixBeforePkgRename?: string;
```

src/changelog.ts
```diff
@@ -264,28 +354,42 @@ export default class Changelog {
...
     tagPrefix = 'v',
     formatter = (changelog) => changelog,
+    versionBeforePkgRename = undefined,
+    tagPrefixBeforePkgRename = undefined,
   }: {
     repoUrl: string;
     tagPrefix?: string;
     formatter?: Formatter;
+    versionBeforePkgRename?: string | undefined;
+    tagPrefixBeforePkgRename?: string | undefined;
```
MajorLift, line 384: Consistent typing for optional values. With `exactOptionalPropertyTypes` disabled, the `| undefined` is redundant.
```suggestion
    versionBeforePkgRename?: string;
    tagPrefixBeforePkgRename?: string;
```

src/changelog.ts
```diff
@@ -264,28 +354,42 @@ export default class Changelog {
...
    * Construct an empty changelog.
    *
    * @param options - Changelog options.
    * @param options.repoUrl - The GitHub repository URL for the current project.
    * @param options.tagPrefix - The prefix used in tags before the version number.
    * @param options.formatter - A function that formats the changelog string.
+   * @param options.versionBeforePkgRename - A version string of the package before being renamed.
+   * An optional, which is required only in case of package renamed.
+   * @param options.tagPrefixBeforePkgRename - A tag prefix string of the package before being renamed.
+   * An optional, which is required only in case of package renamed.
```
MajorLift, line 371: Applying JSDoc syntax for optional params.
```suggestion
   * @param [options.tagPrefix] - The prefix used in tags before the version number.
   * @param [options.formatter] - A function that formats the changelog string.
   * @param [options.versionBeforePkgRename] - A version string of the package before being renamed.
   * An optional, which is required only in case of package renamed.
   * @param [options.tagPrefixBeforePkgRename] - A tag prefix string of the package before being renamed.
   * An optional, which is required only in case of package renamed.
```

src/changelog.ts
```diff
@@ -164,24 +164,19 @@ function getTagUrl(repoUrl: string, tag: string) {
  * @param repoUrl - The URL for the GitHub repository.
  * @param tagPrefix - The prefix used in tags before the version number.
  * @param releases - The releases to generate link definitions for.
+ * @param versionBeforePkgRename - A version string of the package before being renamed.
+ * An optional, which is required only in case of package renamed.
+ * @param tagPrefixBeforePkgRename - A tag prefix string of the package before being renamed.
+ * An optional, which is required only in case of package renamed.
```
MajorLift, line 170:
```suggestion
 * @param [versionBeforePkgRename] - A version string of the package before being renamed.
 * An optional, which is required only in case of package renamed.
 * @param [tagPrefixBeforePkgRename] - A tag prefix string of the package before being renamed.
 * An optional, which is required only in case of package renamed.
```

src/changelog.ts
```diff
@@ -192,42 +187,137 @@ function stringifyLinkReferenceDefinitions(
...
+}
+
+/**
+ * Get a string of unreleased link reference definition.
+ *
+ * @param repoUrl - The URL for the GitHub repository.
+ * @param tagPrefix - The prefix used in tags before the version number.
+ * @param releases - The releases to generate link definitions for.
+ * @param versionBeforePkgRename - A version string of the package before being renamed.
+ * @param tagPrefixBeforePkgRename - A tag prefix string of the package before being renamed.
```
MajorLift, line 223:
```suggestion
 * @param [versionBeforePkgRename] - A version string of the package before being renamed.
 * @param [tagPrefixBeforePkgRename] - A tag prefix string of the package before being renamed.
```

src/changelog.ts
```diff
@@ -192,42 +187,137 @@ function stringifyLinkReferenceDefinitions(
...
+}
+
+/**
+ * Get a list of release link reference definitions.
+ *
+ * @param repoUrl - The URL for the GitHub repository.
+ * @param tagPrefix - The prefix used in tags before the version number.
+ * @param releases - The releases to generate link definitions for.
+ * @param versionBeforePkgRename - A version string of the package before being renamed.
+ * @param tagPrefixBeforePkgRename - A tag prefix string of the package before being renamed.
```
MajorLift, line 266:
```suggestion
 * @param [versionBeforePkgRename] - A version string of the package before being renamed.
 * @param [tagPrefixBeforePkgRename] - A tag prefix string of the package before being renamed.
```

src/cli.ts
```diff
@@ -155,6 +157,10 @@ type ValidateOptions = {
  * @param options.tagPrefix - The prefix used in tags before the version number.
  * @param options.fix - Whether to attempt to fix the changelog or not.
  * @param options.formatter - A custom Markdown formatter to use.
+ * @param options.versionBeforePkgRename - A version of the package before being renamed.
+ * An optional, which is required only in case of package renamed.
+ * @param options.tagPrefixBeforePkgRename - A tag prefix of the package before being renamed.
+ * An optional, which is required only in case of package renamed.
```
MajorLift, line 163:
```suggestion
 * @param [options.versionBeforePkgRename] - A version of the package before being renamed.
 * An optional, which is required only in case of package renamed.
 * @param [options.tagPrefixBeforePkgRename] - A tag prefix of the package before being renamed.
 * An optional, which is required only in case of package renamed.
```

src/cli.ts
MajorLift, line 1:
```diff
@@ -151,6 +151,6 @@ type ValidateOptions = {
  *
  * @param options - Validation options.
  * @param options.changelogPath - The path to the changelog file.
- * @param options.currentVersion - The current project version.
+ * @param [options.currentVersion] - The current project version.
  * @param options.isReleaseCandidate - Whether the current branch is a release candidate or not.
  * @param options.repoUrl - The GitHub repository URL for the current project.
```

MajorLift, line 1:
```diff
@@ -98,10 +98,10 @@ type UpdateOptions = {
  *
  * @param options - Update options.
  * @param options.changelogPath - The path to the changelog file.
- * @param options.currentVersion - The current project version.
+ * @param [options.currentVersion] - The current project version.
  * @param options.isReleaseCandidate - Whether the current branch is a release candidate or not.
  * @param options.repoUrl - The GitHub repository URL for the current project.
- * @param options.projectRootDirectory - The root project directory.
+ * @param [options.projectRootDirectory] - The root project directory.
  * @param options.tagPrefix - The prefix used in tags before the version number.
  * @param options.formatter - A custom Markdown formatter to use.
  */
```

src/parse-changelog.ts
```diff
@@ -32,21 +32,35 @@ function isValidChangeCategory(category: string): category is ChangeCategory {
  * @param options.repoUrl - The GitHub repository URL for the current project.
  * @param options.tagPrefix - The prefix used in tags before the version number.
  * @param options.formatter - A custom Markdown formatter to use.
+ * @param options.versionBeforePkgRename - A version string of the package before being renamed.
+ * An optional, which is required only in case of package renamed.
+ * @param options.tagPrefixBeforePkgRename - A tag prefix string of the package before being renamed.
+ * An optional, which is required only in case of package renamed.
```
MajorLift, line 38:
```suggestion
 * @param [options.tagPrefix] - The prefix used in tags before the version number.
 * @param [options.formatter] - A custom Markdown formatter to use.
 * @param [options.versionBeforePkgRename] - A version string of the package before being renamed.
 * An optional, which is required only in case of package renamed.
 * @param [options.tagPrefixBeforePkgRename] - A tag prefix string of the package before being renamed.
 * An optional, which is required only in case of package renamed.
```

src/validate-changelog.ts
```diff
@@ -90,6 +92,10 @@ type ValidateChangelogOptions = {
  * header, and that there are no unreleased changes present.
  * @param options.tagPrefix - The prefix used in tags before the version number.
  * @param options.formatter - A custom Markdown formatter to use.
+ * @param options.versionBeforePkgRename - A version string of the package before being renamed.
+ * An optional, which is required only in case of package renamed.
+ * @param options.tagPrefixBeforePkgRename - A tag prefix string of the package before being renamed.
+ * An optional, which is required only in case of package renamed.
```
MajorLift, line 98:
```suggestion
 * @param [options.tagPrefix] - The prefix used in tags before the version number.
 * @param [options.formatter] - A custom Markdown formatter to use.
 * @param [options.versionBeforePkgRename] - A version string of the package before being renamed.
 * An optional, which is required only in case of package renamed.
 * @param [options.tagPrefixBeforePkgRename] - A tag prefix string of the package before being renamed.
 * An optional, which is required only in case of package renamed.
```

kanthesha:
@MajorLift I see you have added [] inconsistently around the params and few other changes in couple of places? And if required, it can stay separate, which can be independently reviewed.

MajorLift:
Yes, `[]` is the JSDoc syntax for optional params. They could replace the "An optional, which is required only..." comments although I didn't remove those.

The fixes adding `[]` were cluttering up the feed and there's no variation between them so I thought I'd consolidate them into a single commit.

src/cli.ts
```diff
@@ -257,6 +267,14 @@ function configureCommonCommandOptions(_yargs: Argv) {
       default: 'v',
       description: 'The prefix used in tags before the version number.',
       type: 'string',
+    })
+    .option('versionBeforePkgRename', {
```
mcmire, line 271: See note above abbreviations. I feel like it would be more consistent with how "package" is spelled in this project if the option that this CLI took was `--version-before-package-rename` rather than `--version-before-pkg-rename`. What do you think?

src/changelog.ts
```diff
@@ -192,42 +187,137 @@ function stringifyLinkReferenceDefinitions(
...
+  const unreleasedLinkReferenceDefinition =
+    getUnreleasedLinkReferenceDefinition(
+      repoUrl,
+      tagPrefix,
+      releases,
+      versionBeforePkgRename,
+      tagPrefixBeforePkgRename,
+    );
 
   // The "previous" release that should be used for comparison is not always
```
mcmire, line 199: This comment now feels out of place. The code that refers to the "'previous' release" is now inside of `getReleaseLinkReferenceDefinitions`. Should this comment be located there instead? (Perhaps inside the function, at the top, before everything.)

src/changelog.ts
```diff
@@ -164,24 +164,19 @@ function getTagUrl(repoUrl: string, tag: string) {
...
-    .map(({ version }) => version)
-    .sort((a: Version, b: Version) => {
-      return semver.gt(a, b) ? -1 : 1;
-    });
-  const latestSemverVersion = descendingSemverVersions[0];
-  // A list of release versions in chronological order
-  const chronologicalVersions = releases.map(({ version }) => version);
-  const hasReleases = chronologicalVersions.length > 0;
-
   // The "Unreleased" section represents all changes made since the *highest*
```
mcmire, line 180: This comment now feels slightly out of place. The code that finds the "_highest_ release, not the most recent release" is now inside of `getUnreleasedLinkReferenceDefinition`. What do you think about moving this comment there? (Perhaps inside the function, at the top, before everything.)

src/changelog.ts
```diff
@@ -192,42 +187,137 @@ function stringifyLinkReferenceDefinitions(
...
+    if (version === chronologicalVersions[chronologicalVersions.length - 1]) {
+      diffUrl = getTagUrl(repoUrl, `${tagPrefixToCompare}${version}`);
+    } else {
+      const versionIndex = chronologicalVersions.indexOf(version);
+      const previousVersion = chronologicalVersions
+        .slice(versionIndex)
+        .find((releaseVersion: Version) => {
+          return semver.gt(version, releaseVersion);
+        });
+      // when there is a package renamed below if will fix the validation for renamed package's first release
```
mcmire, line 293: This comment is slightly confusing. Do you mean the first release of the package as a whole, or just the original name of the package?

src/changelog.ts
```diff
@@ -192,42 +187,137 @@ function stringifyLinkReferenceDefinitions(
...
           ? getCompareUrl(
               repoUrl,
-              `${tagPrefix}${previousVersion}`,
-              `${tagPrefix}${version}`,
+              `${tagPrefixToCompare}${previousVersion}`,
+              `${tagPrefixToCompare}${version}`,
             )
-          : getTagUrl(repoUrl, `${tagPrefix}${version}`);
+          : getTagUrl(repoUrl, `${tagPrefixToCompare}${version}`);
       }
```
mcmire, line 315: Two thoughts:

- I'm having a bit of trouble parsing this code. It feels like this ought to be simpler.
- Right now, we check whether this is the first version under the new name of the package (i.e. whether the previous version is the one before the rename); then we check whether there is a previous version at all. It feels like we ought to flip the logic around. 


If we take a step back, we have two conditionals that decide what the URL looks like for a diff:

- Either this is the smallest release, or there is a release before this one.
- Either the package either got renamed (and we know its previous name and the last version it had its previous name), or it did not.

I feel like code would describe these scenarios better:

``` typescript
const versionIndex = chronologicalVersions.indexOf(version);
const previousVersion = chronologicalVersions
  .slice(versionIndex)
  .find((releaseVersion: Version) => {
    return semver.gt(version, releaseVersion);
  });
if (previousVersion) {
  if (
    tagPrefixBeforePkgRename &&
    versionBeforePkgRename === previousVersion
  ) {
    // The package was renamed at this version
    // (the tag prefix holds the new name).
    diffUrl = getCompareUrl(
      repoUrl,
      `${tagPrefixBeforePkgRename}${previousVersion}`,
      `${tagPrefix}${version}`,
    );
  } else {
    // If the package was ever renamed, it was not renamed at this version,
    // so use either the old tag prefix or the new tag prefix.
    // If the package was never renamed, use the tag prefix as it is.
    diffUrl = getCompareUrl(
      repoUrl,
      `${tagPrefixToCompare}${previousVersion}`,
      `${tagPrefixToCompare}${version}`,
    );
  }
} else {
  // This is the smallest release.
  diffUrl = getTagUrl(repoUrl, `${tagPrefixToCompare}${version}`);
}
```

It's also possible I'm a bit hung up because of the name of `tagPrefixToCompare`. This seems to either represent the original or new package name; or just whatever name the package happens to have (if it was not renamed). That said, I am not sure what to call this variable, so perhaps it is good for now.

mcmire (commented):
Sorry for the delay in reviewing. As you can see I left a lot of comments. Try not to be overwhelmed 😅  I think as a whole this implementation makes sense. A lot of the comments are minor.

src/changelog.ts
```diff
@@ -192,42 +187,137 @@ function stringifyLinkReferenceDefinitions(
...
+ * @param [versionBeforePkgRename] - A version string of the package before being renamed.
+ * @param [tagPrefixBeforePkgRename] - A tag prefix string of the package before being renamed.
+ * @returns A unreleased link reference definition string.
+ */
+function getUnreleasedLinkReferenceDefinition(
+  repoUrl: string,
+  tagPrefix: string,
+  releases: ReleaseMetadata[],
+  versionBeforePkgRename?: string,
+  tagPrefixBeforePkgRename?: string,
```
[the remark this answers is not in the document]
MajorLift: This would have the advantage of not breaking if `exactOptionalPropertyTypes` is turned on.

src/changelog.ts
MajorLift: Yes, I like `| undefined` better than `?:` here, since `undefined` is explicitly being passed in with these two variables, and not making them optional will prevent the code from breaking if `exactOptionalPropertyTypes` is turned on. 

My concern is around consistent typing for the other properties in these files that are defined as optional. `exactOptionalPropertyTypes` would enforce that consistency.

src/changelog.ts
```diff
@@ -164,24 +164,19 @@ function getTagUrl(repoUrl: string, tag: string) {
  * @param repoUrl - The URL for the GitHub repository.
  * @param tagPrefix - The prefix used in tags before the version number.
  * @param releases - The releases to generate link definitions for.
+ * @param [versionBeforePkgRename] - A version string of the package before being renamed.
```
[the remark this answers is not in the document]
kanthesha: Make sense. sure.

src/changelog.ts
```diff
@@ -192,42 +187,137 @@ function stringifyLinkReferenceDefinitions(
...
+  tagPrefixBeforePkgRename?: string,
+): string[] {
+  const chronologicalVersions = releases.map(({ version }) => version);
+  let tagPrefixToCompare = tagPrefix;
+  const releaseLinkReferenceDefinitions = releases.map(({ version }) => {
+    let diffUrl;
+    // once the version matches with original version, rest of the lines in changelog will be assumed as migrated tags
+    if (tagPrefixBeforePkgRename && versionBeforePkgRename === version) {
+      tagPrefixToCompare = tagPrefixBeforePkgRename;
+    }
```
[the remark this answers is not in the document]
kanthesha: Sure. As the variables tagPrefixBeforePkgRename versionBeforePkgRename have been renamed. I'll commit this change along with rename.

src/changelog.ts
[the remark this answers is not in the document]
kanthesha: I get what you mean and that's how it is been called now. But when we look at the function utility, it can work even without these parameters. We need these parameters only in case of package is renamed. That way, I think, it's better them to be optional.

src/cli.ts
```diff
@@ -142,6 +142,8 @@ type ValidateOptions = {
   tagPrefix: string;
   fix: boolean;
   formatter: Formatter;
+  versionBeforePkgRename?: string; // used in case of package renamed
+  tagPrefixBeforePkgRename?: string; // used in case of package renamed
```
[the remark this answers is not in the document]
kanthesha: I think, the optional parameters will allow the consumer of the function to not to worry about passing those params or what to pass which are not relevant for them.

src/changelog.ts
kanthesha: >renamed package's first release
the first release after the package has been renamed. I'll update it.

src/changelog.ts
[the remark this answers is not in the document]
mcmire: Alright. I guess the benefit of defining the variable inside the loop is that it creates a smaller chunk of code that can be understood on its own and refactored more easily in the future if necessary. This is minor, however.

src/changelog.ts
MajorLift: @kanthesha I think you're right that this could be done in its own PR.

src/changelog.ts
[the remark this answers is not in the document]
mcmire: Ah. I see what you're saying.

What do you think about using a `for` loop rather than `map` then? I feel like this fact — that we _want_ a variable whose state persists outside of the loop rather than getting reset — is important and shouldn't get lost. By including it in the `map` I wonder if it would get glossed over, since the primary goal of the `map` is to build the string. But if we use a `for` loop we're intentionally using a more procedural pattern and thus we're communicating that there are multiple things we want the reader to pay attention to.

src/changelog.ts
MajorLift: Moved to https://github.com/MetaMask/auto-changelog/pull/158

src/cli.ts
MajorLift: Moved to https://github.com/MetaMask/auto-changelog/pull/159

MajorLift: Moved to https://github.com/MetaMask/auto-changelog/pull/159

src/parse-changelog.ts
```diff
@@ -30,23 +30,37 @@ function isValidChangeCategory(category: string): category is ChangeCategory {
...
   changelogContent: string;
   repoUrl: string;
   tagPrefix?: string;
   formatter?: Formatter;
+  versionBeforePackageRename?: string;
+  tagPrefixBeforePackageRename?: string;
 }) {
   const changelogLines = changelogContent.split('\n');
-  const changelog = new Changelog({ repoUrl, tagPrefix, formatter });
+  const changelog = new Changelog({
```
mcmire, line 57: Since it would now be possible to change the behavior by passing these options, what do you think about adding a new test which verifies that a changelog that features renamed tags can be successfully parsed?

mcmire (commented):
I have another suggestion for how we can make the code match the intended usage.

src/changelog.ts
```diff
@@ -164,24 +164,57 @@ function getTagUrl(repoUrl: string, tag: string) {
  * @param repoUrl - The URL for the GitHub repository.
  * @param tagPrefix - The prefix used in tags before the version number.
  * @param releases - The releases to generate link definitions for.
+ * @param [versionBeforePackageRename] - A version string of the package before being renamed.
```
mcmire, line 167: Sorry, I think I missed this when the suggestion was made. I don't think we should have to provide brackets around arguments in TypeScript files. This is required when using JSDoc, but neither VSCode nor `typedoc` honors this syntax when interpreting JSDoc blocks, and we also have TypeScript types to indicate when something is optional. See comment here: https://github.com/MetaMask/auto-changelog/pull/159#pullrequestreview-1680555339. Thoughts on removing this (and other places where this has been added)?

MajorLift: Reverted https://github.com/MetaMask/auto-changelog/commit/b51a4fa27dff4b46f3452e6b799a3d3c014c06bb as per discussion in https://github.com/MetaMask/auto-changelog/pull/159#pullrequestreview-1680555339.

src/cli.ts
mcmire, line 1: Judging by how we're taking in command-line arguments, it looks like it is possible for `versionBeforePackageRename` and `tagPrefixBeforePackageRename` to not be supplied together. I would imagine we probably don't want this.

It looks like we are already making some checks around command-line arguments ending on line 435. What do you think about adding a couple more if statements here?

``` typescript
if (
  (versionBeforePackageRename && !tagPrefixBeforePackageRename) ||
  (!versionBeforePackageRename && tagPrefixBeforePackageRename)
) {
  return exitWithError(
    "--version-before-package-rename and --tag-prefix-before-package-rename must be given together or not at all.',
  );
}
```

src/changelog.ts
```diff
@@ -164,24 +164,57 @@ function getTagUrl(repoUrl: string, tag: string) {
...
+ * An optional, which is required only in case of package renamed.
+ * @param [tagPrefixBeforePackageRename] - A tag prefix string of the package before being renamed.
+ * An optional, which is required only in case of package renamed.
  * @returns The stringified release link definitions.
  */
 function stringifyLinkReferenceDefinitions(
   repoUrl: string,
   tagPrefix: string,
   releases: ReleaseMetadata[],
+  versionBeforePackageRename?: string,
```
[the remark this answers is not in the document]
mcmire: We could do that. One could argue that `changelog.ts` should be concerned primarily with defining the Changelog class, so what do you think about creating a new file to hold shared types like this?

src/parse-changelog.ts
mcmire: It looks like the test you pointed to covers the changes you've made to `validateChangelog`. However, `parseChangelog` is used in both `validateChangelog` and `updateChangelog`, and so I think it'd be valuable for us to confirm that it works well when called on its own. Does that make sense?

src/changelog-config.ts
mcmire: Resolving this conversation since this code was removed.

src/changelog.ts
```diff
@@ -259,32 +252,26 @@ function getUnreleasedLinkReferenceDefinition(
...
   const releaseLinkReferenceDefinitions = releases.map(({ version }) => {
     let diffUrl;
-    // once the version matches with versionBeforePackageRename, rest of the lines in changelog will be assumed as migrated tags
-    if (
-      tagPrefixBeforePackageRename &&
-      versionBeforePackageRename === version
-    ) {
-      tagPrefixToCompare = tagPrefixBeforePackageRename;
+    // once the version matches with versionBeforeRename, rest of the lines in changelog will be assumed as migrated tags
+    if (packageRename && packageRename?.versionBeforeRename === version) {
```
mcmire, line 274: Nit: Since you're already checking for `packageRename` once, does it make sense to drop the extra check?

``` suggestion
    if (packageRename && packageRename.versionBeforeRename === version) {

File: src/changelog.ts [hunk 35]
```diff
@@ -364,42 +351,35 @@ export default class Changelog {
...
    * @param options - Changelog options.
    * @param options.repoUrl - The GitHub repository URL for the current project.
-   * @param options.tagPrefix - The prefix used in tags before the version number.
-   * @param options.formatter - A function that formats the changelog string.
-   * @param options.versionBeforePackageRename - A version string of the package before being renamed.
-   * An optional, which is required only in case of package renamed.
-   * @param options.tagPrefixBeforePackageRename - A tag prefix string of the package before being renamed.
+   * @param [options.tagPrefix] - The prefix used in tags before the version number.
+   * @param [options.formatter] - A function that formats the changelog string.
+   * @param [options.packageRename] - The package rename properties.
```
mcmire, line 364: @MajorLift and I had a discussion about using brackets to indicate optional arguments [here](https://github.com/MetaMask/auto-changelog/pull/159), and we decided that they aren't necessary after all, since both VSCode and `typedoc` both ignore these, and since we already use TypeScript types to communicate optionality. For consistency with our libraries would you mind removing these? Sorry for the back and forth on this.

``` suggestion
   * @param options.tagPrefix - The prefix used in tags before the version number.
   * @param options.formatter - A function that formats the changelog string.
   * @param options.packageRename - The package rename properties.

mcmire (commented):
I have one more suggestion around removing the brackets for the optional arguments, but this otherwise looks good to me.

mcmire (approved):
Looks good to me!

PR #158, opened 2023-10-13
Remove `as` assertions, or fix to mitigate the risk of silent failure

MajorLift:
## Explanation

- [Replace `as T[]` assertions with `?? []`](https://github.com/MetaMask/auto-changelog/commit/53cf6735648f216ce3c8c5fe94f62cf078a63f2c) 

`as` assertions intended to exclude an empty/nullable type can be replaced by a nullish coalescing operator converting nullable values into an acceptable empty value that doesn't pollute the variable's type signature.

- [Replace `as` assertions with redundant type guards](https://github.com/MetaMask/auto-changelog/commit/f950f3e43d90403068c672eb00cbaa378bc77866) 

Sometimes TypeScript's type narrowing and inference doesn't fully work quite like it should. Even in these cases, using `as` assertions should be avoided.

It may seem like we know for certain what the type should be and there's no risk of it breaking, but there's always the possibility that some combination of changes elsewhere might affect the typing, in which case `as` will suppress the alerts TypeScript would otherwise provide us with. This may lead to code breaking silently.

In this example, adding type guards and null checks -- even if they're redundant -- is a small cost to avoid silent failure.

- [Make `as` assertion rely on inference instead of hard-coded type casting](https://github.com/MetaMask/auto-changelog/commit/8da9f215a13db6775d178968ff343144fc3f7ffb) 

If anything changes in the typing of `ChangeCategory` or `unreleasedChanges` that make them inconsistent, this code will fail silently. Relying on type inference instead of explicit type casting will make this logic a little less brittle.

## References

- See https://github.com/MetaMask/contributor-docs/issues/47

### DIFF
--- a/src/changelog.ts
+++ b/src/changelog.ts
@@ -1,3 +1,4 @@
+import { getKnownPropertyNames } from '@metamask/utils';
 import semver from 'semver';
 
 import {
@@ -383,9 +384,7 @@ export default class Changelog {
 
     const unreleasedChanges = this.#changes[unreleased];
 
-    for (const category of Object.keys(
-      unreleasedChanges,
-    ) as (keyof typeof unreleasedChanges)[]) {
+    for (const category of getKnownPropertyNames(unreleasedChanges)) {
       if (releaseChanges[category]) {
         releaseChanges[category] = [
           ...(unreleasedChanges[category] ?? []),
--- a/src/changelog.ts
+++ b/src/changelog.ts
@@ -89,7 +89,7 @@ function stringifyRelease(
   const categorizedChanges = orderedChangeCategories
     .filter((category) => categories[category])
     .map((category) => {
-      const changes = categories[category] as string[];
+      const changes = categories[category] ?? [];
       return stringifyCategory(category, changes);
     })
     .join('\n\n');
@@ -386,8 +386,8 @@ export default class Changelog {
     for (const category of Object.keys(unreleasedChanges) as ChangeCategory[]) {
       if (releaseChanges[category]) {
         releaseChanges[category] = [
-          ...(unreleasedChanges[category] as string[]),
-          ...(releaseChanges[category] as string[]),
+          ...(unreleasedChanges[category] ?? []),
+          ...(releaseChanges[category] ?? []),
         ];
       } else {
         releaseChanges[category] = unreleasedChanges[category];
--- a/src/changelog.ts
+++ b/src/changelog.ts
@@ -383,7 +383,9 @@ export default class Changelog {
 
     const unreleasedChanges = this.#changes[unreleased];
 
-    for (const category of Object.keys(unreleasedChanges) as ChangeCategory[]) {
+    for (const category of Object.keys(
+      unreleasedChanges,
+    ) as (keyof typeof unreleasedChanges)[]) {
       if (releaseChanges[category]) {
         releaseChanges[category] = [
           ...(unreleasedChanges[category] ?? []),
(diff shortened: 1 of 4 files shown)


src/changelog.ts
```diff
@@ -383,11 +383,13 @@ export default class Changelog {
...
-    for (const category of Object.keys(unreleasedChanges) as ChangeCategory[]) {
+    for (const category of Object.keys(
+      unreleasedChanges,
+    ) as (keyof typeof unreleasedChanges)[]) {
       if (releaseChanges[category]) {
         releaseChanges[category] = [
-          ...(unreleasedChanges[category] as string[]),
-          ...(releaseChanges[category] as string[]),
+          ...(unreleasedChanges[category] ?? []),
+          ...(releaseChanges[category] ?? []),
```
mcmire, line 392: `@metamask/utils` now has a `getKnownPropertyNames` function which is specifically designed to iterate over the properties of an enum, while retaining types. This should obviate the need for the type assertion here at least. Note that this requires adding `@metamask/utils` as a dependency, but I think that's okay (we use it everywhere):

``` suggestion
    for (const category of getKnownPropertyNames(unreleasedChanges)) {
      if (releaseChanges[category]) {
        releaseChanges[category] = [
          ...(unreleasedChanges[category] ?? []),
          ...(releaseChanges[category] ?? []),

File: src/update-changelog.ts [hunk 2]
```diff
@@ -250,19 +250,16 @@ export async function updateChangelog({
...
     isReleaseCandidate &&
+    currentVersion &&
     !changelog
       .getReleases()
       .find((release) => release.version === currentVersion)
   ) {
-    // Typecast: currentVersion will be defined here due to type guard at the
-    // top of this function.
-    changelog.addRelease({ version: currentVersion as Version });
+    changelog.addRelease({ version: currentVersion });
```
mcmire, line 258: This comment is referring to this code at the top of the function:

``` typescript
if (isReleaseCandidate && !currentVersion) {
  throw new Error(
    `A version must be specified if 'isReleaseCandidate' is set.`,
  );
}
```

Since we're checking for `isReleaseCandidate` here, there's no way that `currentVersion` is not set. Hence why I believe that the typecast was added originally. (Why TypeScript doesn't see this, I'm not sure.)

This change doesn't hurt, but I wanted to make sure you saw that.

mcmire (approved):
One suggestion here that could remove all of the typecasts, but it is non-blocking. Thanks for doing this.

src/changelog.ts
MajorLift: This is really neat! Thanks for the pointer.

src/update-changelog.ts
```diff
@@ -247,38 +233,47 @@ export async function updateChangelog({
...
+    }
 
-  if (isReleaseCandidate && hasUnreleasedChanges) {
-    // Typecast: currentVersion will be defined here due to type guard at the
-    // top of this function.
-    changelog.migrateUnreleasedChangesToRelease(currentVersion as Version);
-  }
+    if (mostRecentTag === `${tagPrefixes[0]}${currentVersion ?? ''}`) {
+      throw new Error(
+        `Current version already has tag, which is unexpected for a release candidate.`,
```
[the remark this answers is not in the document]
MajorLift: @kanthesha Let me know if https://github.com/MetaMask/auto-changelog/pull/158/commits/a26b0bbf0a6222b4bf2e1dbb7dbd74a725679bd0 looks good!

mcmire (approved):
LGTM!

PR #159, opened 2023-10-13
Apply `[]` syntax for optional params to JSDoc `@param` tags

MajorLift:
- It can be tricky to tell whether or not a parameter is optional if they are passed in as object properties that might also be nested.
- The JSDoc `[]` syntax gives a simple but clear visual indication of which params are optional.
- This PR annotates all existing params that are defined using the `?:` operator.

### DIFF
--- a/src/changelog.ts
+++ b/src/changelog.ts
@@ -74,8 +74,8 @@ function stringifyCategory(category: ChangeCategory, changes: string[]) {
  * @param version - The release version.
  * @param categories - The categories of changes included in this release.
  * @param options - Additional release options.
- * @param options.date - The date of the release.
- * @param options.status - The status of the release (e.g., "DEPRECATED").
+ * @param [options.date] - The date of the release.
+ * @param [options.status] - The status of the release (e.g., "DEPRECATED").
  * @returns The stringified release section.
  */
 function stringifyRelease(
@@ -269,8 +269,8 @@ export default class Changelog {
    *
    * @param options - Changelog options.
    * @param options.repoUrl - The GitHub repository URL for the current project.
-   * @param options.tagPrefix - The prefix used in tags before the version number.
-   * @param options.formatter - A function that formats the changelog string.
+   * @param [options.tagPrefix] - The prefix used in tags before the version number.
+   * @param [options.formatter] - A function that formats the changelog string.
    */
   constructor({
     repoUrl,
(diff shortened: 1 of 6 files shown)


mcmire (commented):
I noticed you suggested these changes on https://github.com/MetaMask/auto-changelog/pull/157. In the extension and mobile app you may see this technique, but that's because most of the files are still written using JavaScript. However, in projects that use TypeScript, we've stopped using it, because now any information related to types can be gleaned from TypeScript.

This is further explained by looking at where JSDoc is used:

1. When writing code (i.e. hovering on a symbol). In this case, VSCode seems to ignore brackets around arguments. If you hover over a function, it doesn't show types for arguments at all; if you hover over an argument, it'll show you the type (which you specify separately).
2. When viewing generated documentation. We use `typedoc` to generate documentation, and [this tool does not read the brackets around arguments](https://typedoc.org/tags/param/).
3. When reading code. If you want to know whether an argument is optional, you can look at the TypeScript types, you don't need to read the JSDoc block to determine this.

Therefore, I'm not sure that we need to add these changes. Thoughts?

MajorLift:
I think the brackets might still be useful when function params are passed in as objects, and those types have nested optional properties.  

For example, if we were to add a nested entry into `ValidateChangeLogOptions` like `packageRename?: PackageRenameOptions` and some properties of `PackageRenameOptions` were optional, JSDoc `[]` syntax highlighting makes it easy to identify the optional arguments at a glance, without having to click and hover our way through multiple types.

I guess the utility of doing this for all typescript projects is limited, though, if the language server and docgen don't utilize the brackets. Just stating "optional" in human language like we're doing now might be good enough.

mcmire:
Yeah. I mean, it's a good thought.

If you keep running into this and still think it would be valuable, it would be nice if we could find or add an ESLint rule around this so it could automatically add the brackets and they would be present without us having to think about it.

MajorLift:
It seems like there are no eslint rules for optional `@param` or `@property` tags, at least in `eslint-plugin-jsdoc`. I guess we could add the brackets if we come across a situation where they'd be especially useful, without enforcing consistent usage everywhere.

Closing this PR for now. Will revert https://github.com/MetaMask/auto-changelog/pull/157/commits/b51a4fa27dff4b46f3452e6b799a3d3c014c06bb as well.

PR #170, opened 2023-10-24
Avoid circular dependency on @metamask/utils

mcmire:
`@metamask/utils` has a development dependency on this package (since we need a way to validate its changelog). So, if this package depends on `@metamask/utils`, then that creates a circular dependency which could potentially cause issues. Since we only use one function from this package, we can just copy it.

---

See: https://github.com/MetaMask/utils/pull/153#issuecomment-1776854260

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
     "changelog": "node dist/cli.js"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


Mrtenz (approved):

cryptodev-2s (approved):
LGTM!

PR #181, opened 2023-12-08
Fix bug introduced to `updateChangelog` in #158

### LINKED ISSUE #180: `updateChangelog` does nothing if `isReleaseCandidate` is `false`
`updateChangelog` seems to have been changed in https://github.com/MetaMask/auto-changelog/pull/158 such that it does nothing if `isReleaseCandidate` is `false`.

Curiously, these changes did not cause the tests to fail. That could indicate some missing test cases. We should add the missing tests and revert the behavior. It would also be helpful to go back and update the changelog entry for this PR and indicate that the change was breaking.

MajorLift:
## Motivation

#158 incorrectly refactored the branches in updateChangelog, resulting in a state where new changelog entries weren't added to the return value if isReleaseCandidate is false.

## Explanation

- This error is fixed by moving the logic for retrieving new entries out of the if (isReleaseCandidate) block.
- The code for retrieving new entries is also extracted into its own method: getNewChangeEntries, abstracting away details that obscured the flow of the method's core logic.
- To verify that the buggy logic is correctly restored, it's necessary to compare the current state of updateChangelog to its state before the bug was introduced. For this, see: [diff link](https://github.com/MetaMask/auto-changelog/compare/e8df1ec717f534c8fe84c46ea86a847fa5a32973..da39a58a55571cf4e8a03e28096aa12c02c75f77#diff-4228e8302e41dd1c51e813af8368efdcb40b3d9db3e3f21f417ae87275d9f389R249-R316).

## References

- Closes #180
- Followed by #188

### DIFF
--- a/src/update-changelog.ts
+++ b/src/update-changelog.ts
@@ -310,11 +310,10 @@ export async function updateChangelog({
       description,
     });
   }
-  const newChangelogContent = changelog.toString();
 
-  return changelogContent === newChangelogContent
-    ? undefined
-    : newChangelogContent;
+  const newChangelogContent = changelog.toString();
+  const isChangelogUpdated = changelogContent !== newChangelogContent;
+  return isChangelogUpdated ? newChangelogContent : undefined;
 }
 
 /**
--- a/src/update-changelog.ts
+++ b/src/update-changelog.ts
@@ -157,6 +157,54 @@ async function getCommitHashesInRange(
   return await runCommand('git', revListArgs);
 }
 
+type AddNewCommitsOptions = {
+  mostRecentTag: string | null;
+  repoUrl: string;
+  loggedPrNumbers: string[];
+  projectRootDirectory?: string;
+};
+
+/**
+ * Get the list of new change entries to add to a changelog.
+ *
+ * @param options - Options.
+ * @param options.mostRecentTag - The most recent tag.
+ * @param options.repoUrl - The GitHub repository URL for the current project.
+ * @param options.loggedPrNumbers - A list of all pull request numbers included in the relevant parsed changelog.
+ * @param options.projectRootDirectory - The root project directory, used to
+ * filter results from various git commands. This path is assumed to be either
+ * absolute, or relative to the current directory. Defaults to the root of the
+ * current git repository.
+ * @returns A list of new change entries to add to the changelog, based on commits made since the last release.
+ */
+async function getNewChangeEntries({
+  mostRecentTag,
+  repoUrl,
+  loggedPrNumbers,
+  projectRootDirectory,
+}: AddNewCommitsOptions) {
+  const commitRange =
+    mostRecentTag === null ? 'HEAD' : `${mostRecentTag}..HEAD`;
+  const commitsHashesSinceLastRelease = await getCommitHashesInRange(
+    commitRange,
+    projectRootDirectory,
+  );
+  const commits = await getCommits(commitsHashesSinceLastRelease);
+
+  const newCommits = commits.filter(
+    ({ prNumber }) =>
+      prNumber === undefined || !loggedPrNumbers.includes(prNumber),
+  );
+
+  return newCommits.map(({ prNumber, description }) => {
+    if (prNumber) {
+      const suffix = `([#${prNumber}](${repoUrl}/pull/${prNumber}))`;
+      return `${description} ${suffix}`;
+    }
+    return description;
+  });
+}
+
 export type UpdateChangelogOptions = {
   changelogContent: string;
   currentVersion?: Version;
@@ -205,11 +253,6 @@ export async function updateChangelog({
   formatter = undefined,
   packageRename,
 }: UpdateChangelogOptions): Promise<string | undefined> {
-  if (isReleaseCandidate && !currentVersion) {
-    throw new Error(
-      `A version must be specified if 'isReleaseCandidate' is set.`,
-    );
-  }
   const changelog = parseChangelog({
     changelogContent,
     repoUrl,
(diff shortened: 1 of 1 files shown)


src/update-changelog.ts
```diff
@@ -213,43 +264,17 @@ export async function updateChangelog({
...
-    commitRange,
-    projectRootDirectory,
-  );
-  const commits = await getCommits(commitsHashesSinceLastRelease);
-
-  const loggedPrNumbers = getAllLoggedPrNumbers(changelog);
-  const newCommits = commits.filter(
-    ({ prNumber }) =>
-      prNumber === undefined || !loggedPrNumbers.includes(prNumber),
-  );
```
MajorLift, line 234: This section is moved to `getNewChangeEntries`

src/update-changelog.ts
```diff
@@ -213,43 +264,17 @@ export async function updateChangelog({
     packageRename,
   });
 
-  // Ensure we have all tags on remote
-  await runCommand('git', ['fetch', '--tags']);
```
MajorLift, line 217: This line is moved to `getMostRecentTag`

src/update-changelog.ts
```diff
@@ -264,28 +289,31 @@ export async function updateChangelog({
...
+      version: isReleaseCandidate ? currentVersion : undefined,
+      category: ChangeCategory.Uncategorized,
+      description,
+    });
   }
 
-  return changelog.toString();
+  const newChangelogContent = changelog.toString();
+  const isChangelogUpdated = changelogContent !== newChangelogContent;
+  return isChangelogUpdated ? newChangelogContent : undefined;
```
MajorLift, line 316: `update()` in `cli.ts` expects that `updateChangelog` return `undefined` if there are no updates to be added to the changelog.

src/update-changelog.ts
```diff
@@ -264,28 +289,31 @@ export async function updateChangelog({
...
     }
+  }
 
-    const newChangeEntries = newCommits.map(({ prNumber, description }) => {
-      if (prNumber) {
-        const suffix = `([#${prNumber}](${repoUrl}/pull/${prNumber}))`;
-        return `${description} ${suffix}`;
-      }
-      return description;
-    });
```
MajorLift, line 277: This section is moved into `getNewChangeEntries`

src/update-changelog.ts
MajorLift, line 234: This section is moved to `getNewChangeEntries`.

src/update-changelog.ts
MajorLift, line 277: This section is moved to `getNewChangeEntries`.

src/update-changelog.ts
MajorLift, line 217: This line is moved to `getMostRecentTag()`.

src/update-changelog.ts
```diff
@@ -213,43 +264,17 @@ export async function updateChangelog({
...
-
   if (isReleaseCandidate) {
     if (!currentVersion) {
       throw new Error(
         `A version must be specified if 'isReleaseCandidate' is set.`,
       );
     }
-
-    if (mostRecentTag === `${tagPrefixes[0]}${currentVersion ?? ''}`) {
+    if (mostRecentTag === `${tagPrefixes[0]}${currentVersion}`) {
```
MajorLift, line 277: `currentVersion` is narrowed to `string` thanks to the `if (!currentVersion)` block being moved directly above. Previously `as string` or `?? ''` was required.

src/update-changelog.ts
```diff
@@ -264,28 +289,31 @@ export async function updateChangelog({
       changelog.addRelease({ version: currentVersion });
```
MajorLift, line 289: `currentVersion` is correctly narrowed due to all related branches being refactored to fall under the `if (isReleaseCandidate)` block instead of there being multiple `if (isReleaseCandidate &&)` blocks. `currentVersion as Version` casting is now unnecessary.

src/update-changelog.ts
```diff
@@ -264,28 +289,31 @@ export async function updateChangelog({
...
+    projectRootDirectory,
+  });
 
-    for (const description of newChangeEntries.reverse()) {
-      changelog.addChange({
-        version: isReleaseCandidate ? currentVersion : undefined,
-        category: ChangeCategory.Uncategorized,
-        description,
-      });
-    }
```
MajorLift, line 285: This is the bracket that caused the bug. Moving it to above the `getNewChangeEntries` call resolves the error.

src/update-changelog.ts
```diff
@@ -264,28 +289,31 @@ export async function updateChangelog({
...
+  for (const description of newChangeEntries.reverse()) {
+    changelog.addChange({
+      version: isReleaseCandidate ? currentVersion : undefined,
+      category: ChangeCategory.Uncategorized,
+      description,
+    });
   }
```
MajorLift, line 312: This is the bracket that caused the bug. Moving it to above the `getNewChangeEntries` call resolves the error.

src/update-changelog.ts
```diff
@@ -157,6 +160,54 @@ async function getCommitHashesInRange(
...
+ * @returns A list of new change entries to add to the changelog, based on commits made since the last release.
+ */
+async function getNewChangeEntries({
+  mostRecentTag,
+  repoUrl,
+  loggedPrNumbers,
+  projectRootDirectory,
+}: AddNewCommitsOptions) {
+  const commitRange =
+    mostRecentTag === null ? 'HEAD' : `${mostRecentTag}..HEAD`;
```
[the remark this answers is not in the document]
MajorLift: This results in a linter error: 
```
Unexpected negated condition. eslint(no-negated-condition)
```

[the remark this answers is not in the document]
mcmire: Hmm, we did use `null` intentionally to represent lack of a most recent tag, so the use of `null` matches the type that `getMostRecentTag` returns. However, I suppose that the tag name could be an empty string in theory if for some reason `git describe` returns an empty string, and that case we would want to fall back to `HEAD`. So, we could update `getMostRecentTag` to just return `string` instead of `string | null`, except that in case there is no recent tag, it would then `return '';`. That might look strange, but we could add a comment saying why we're doing that or give that empty string a name that describe its purpose.

In any case, it seems that this line was merely copied from another function, so maybe we want to make that change in another PR?

src/update-changelog.ts
```diff
@@ -213,43 +263,17 @@ export async function updateChangelog({
...
-
   if (isReleaseCandidate) {
     if (!currentVersion) {
       throw new Error(
         `A version must be specified if 'isReleaseCandidate' is set.`,
       );
     }
-
-    if (mostRecentTag === `${tagPrefixes[0]}${currentVersion ?? ''}`) {
+    if (mostRecentTag === `${tagPrefixes[0]}${currentVersion}`) {
```
mcmire, line 276: Hmm, I just realized that we only use the first tag prefix here, which seems strange. This may be a potential bug. But we can investigate that later.

mcmire (approved):
Hopefully in writing the tests no bugs are revealed, but comparing this to the original version, this makes sense to me.

src/update-changelog.ts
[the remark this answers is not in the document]
MajorLift: I like using `null` for clear distinction and visibility, but in any case I agree with revisiting this in its own ticket since this wasn't a change introduced by this PR.

src/update-changelog.ts
MajorLift: The jsdoc does specify that the first prefix is the intended prefix and the rest are fallbacks. `parseChangelog` also only uses the first prefix. 

`getMostRecentTag` does iterate over the tagPrefix list. It should return the used prefix, or we could extract the prefix from the returned most recent tag.

But this also wasn't a change introduced in this PR, so it should get its own ticket.

PR #184, opened 2023-12-19
Add changelog update usage for package rename options in README

cryptodev-2s:
Update the CLI Usage of `auto-changelog update` for renamed packages in the README to mention `--tag-prefix-before-package-name` and `--version-before-package-rename`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,18 @@ or
 
 `npm run auto-changelog update --rc`
 
+#### Update the changelog for a renamed package
+
+This option is designed to be used for packages that live in a monorepo.
+
+For instance, if your package is called `polling-controller` and was renamed to `@metamask/polling-controller` at version 0.2.3, and thus the Git tags followed suit:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,18 @@ or
 
 `npm run auto-changelog update --rc`
 
+#### Update the changelog for a renamed package
+
+This option is designed to be used for packages that live in a monorepo.
```
mikesposito, line 37: Can this also apply to a package that is not in a monorepo, but that has just been renamed?

cryptodev-2s: @mikesposito actually I got inspired by this PR merged some time ago https://github.com/MetaMask/auto-changelog/pull/173 ? I am not sure why this won't apply for other than monorepos but my guess it should

mikesposito: Oh, I see! Thanks for linking that.

As per our naming convention, release tags for packages inside a monorepo have a "tag prefix", and that tag prefix is used when building links in the changelog. As our single packages just use version as a tag, they don't need any tweak on the prefix - as they don't have it.

mikesposito (approved):
Looks good!

PR #199, opened 2024-08-13
feat(getMostRecentTags): add fetchRemote option to allow disabling remote fetching

legobeat:
Currently, the `getMostRecentTags` will perform a `git fetch --tags`, failing if the remote fetching does.

This adds a new option `fetchRemote` (default `true`). Setting it to `false` disables the remote-fetching.

Motivation: Allow running https://github.com/MetaMask/create-release-branch/ in an unauthenticated context.

### DIFF
--- a/src/update-changelog.ts
+++ b/src/update-changelog.ts
@@ -11,17 +11,25 @@ import { PackageRename } from './shared-types';
  * Get the most recent tag for a project.
  *
  * @param options - Options.
+ * @param options.fetchRemote - Whether to synchronize local tags with remote.
  * @param options.tagPrefixes - A list of tag prefixes to look for, where the first is the intended
  * prefix and each subsequent prefix is a fallback in case the previous tag prefixes are not found.
+ * @param options.run - Optional alternative shell command interpreter
  * @returns The most recent tag.
  */
 async function getMostRecentTag({
+  fetchRemote,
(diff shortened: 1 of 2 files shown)


src/update-changelog.ts
```diff
@@ -11,17 +11,22 @@ import { PackageRename } from './shared-types';
...
   tagPrefixes,
 }: {
+  fetchRemote?: boolean;
   tagPrefixes: [string, ...string[]];
 }) {
   // Ensure we have all tags on remote
-  await runCommand('git', ['fetch', '--tags']);
+  if (typeof fetchRemote !== 'boolean' || fetchRemote) {
+    await runCommand('git', ['fetch', '--tags']);
+  }
```
mikesposito, line 29: Could we cover this with some tests?

legobeat: This entire module is currently untested, unfortunately...

legobeat: basic unit test for the new behavior has been added.

mcmire:
Coming back to this PR. I think this would be useful to have, although for an additional reason than originally proposed. It appears that both `create-release-branch` and `auto-changelog` fetch tags, which is not necessary. Adding a flag to `auto-changelog` would be useful so that `create-release-branch` doesn't end up fetching tags twice.

PR #202, opened 2024-09-23
BREAKING: Bump `prettier` to `^3.3.3`

### LINKED ISSUE #201: Bump Prettier to v3
We should update Prettier to v3, and fix any breaking changes. Prettier v3 uses an async API, meaning that we need to update the API of this package to be async as well.

Mrtenz:
This bumps Prettier to `^3.3.3`, and fixes any breaking changes in Prettier.

Closes #201.

## Breaking changes

- `prettier@^3.3.3` is now a peer-dependency of this package.
- `Changelog.toString` now returns a `Promise`, which resolves with the formatted changelog.


Mrtenz:
@SocketSecurity ignore npm/create-jest@29.7.0

New author ok.

@SocketSecurity ignore npm/jake@10.9.2

Shell access expected.

package.json
```diff
@@ -52,22 +44,34 @@
...
+    "prettier": "^3.3.3",
+    "prettier-plugin-packagejson": "^2.5.2",
     "rimraf": "^3.0.2",
-    "ts-jest": "^26.5.6",
+    "ts-jest": "^29.2.5",
     "typescript": "~4.8.4"
   },
+  "peerDependencies": {
+    "prettier": "^3.3.3"
+  },
```
mikesposito, line 63: This forces a consumer to use this specific version range of `prettier` - curious, is there any unwanted outcome of including this in direct dependencies so that this package can use its specific version regardless of the rest of the dep tree?

package.json
```diff
@@ -52,22 +44,34 @@
...
+  "packageManager": "yarn@3.2.4",
+  "engines": {
+    "node": ">=14.0.0"
+  },
```
mikesposito, line 67: Supported versions were `^18.18 || >=20`, any reason we need v14?

Mrtenz: I messed this up after rebasing. Fixed it now.

package.json
Mrtenz: I assume every consumer already has Prettier installed. We could change it to a version range like `>=3.0.0`.

src/validate-changelog.ts
```diff
@@ -107,7 +107,7 @@ type ValidateChangelogOptions = {
  * `true` and the changelog contains uncategorized changes.
  * @throws `ChangelogFormattingError` - Will throw if there is a formatting error.
  */
-export function validateChangelog({
+export async function validateChangelog({
```
mikesposito, line 110: Nit:  `README` should be updated with the new API (though, I believe it should be updated with other things as well i.e. node version)

mikesposito (approved):
Looks great!

src/validate-changelog.ts
Mrtenz: Will do in a separate PR.

PR #212, opened 2025-01-21
Auto categorize CLI flag

jake-perkins:
This adds a new --autoCategorize flag to the CLI which will categorize the conventional commits to keep a changelog format for the generated changelog

Example output runs

yarn auto-changelog update --autoCategorize

---

### Uncategorized

- tweaking configs
- newline
- update changelog
- remove logs
- testing
- reset changelog
- readme
- format changelog

### Added

- feat: testing conv commit feat
- feat: categorize conventional commits

### Fixed

- fix: format
- fix: testing conv commit


sethkfman (dismissed):
LGTM

src/update-changelog.ts
```diff
@@ -215,6 +215,7 @@ export type UpdateChangelogOptions = {
   projectRootDirectory?: string;
   tagPrefixes?: [string, ...string[]];
   formatter?: Formatter;
+  autoCategorize: boolean;
```
cryptodev-2s, line 218: Nit: should this be optional

mcmire (commented):
This makes sense to me, I just had some suggestions.

It doesn't appear that we have any tests for `updateChangelog`. However, would you mind adding some in an `update-changelog.test.ts` file for your new changes? It doesn't have to test any of the existing functionality, just your addition.

src/update-changelog.ts
```diff
@@ -329,3 +337,27 @@ async function runCommand(command: string, args: string[]): Promise<string[]> {
...
+ *
+ * @param description - The commit message description.
+ * @returns The category of the change.
+ */
+function getCategory(description: string): ChangeCategory {
+  // Check if description contains a colon
+  if (description.includes(':')) {
+    const [prefix] = description.split(':').map((part) => part.trim());
+    switch (prefix) {
+      case 'feat':
```
mcmire, line 352: Nit: Should we extract these prefixes to an enum so that they are more easily referenceable?

README.md
```diff
@@ -24,6 +24,10 @@ or
 
 `npm run auto-changelog update`
 
+#### Generate Categorized sections of the changelog
+
+`yarn run auto-changelog update --autoCategorize`
```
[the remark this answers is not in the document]
jake-perkins: Let me know your preference, I saw we had a few different conventions ( currentVersion ) was another example argument. happy to update it to whatever you prefer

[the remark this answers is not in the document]
mcmire: Ah, so we do... hmm. I guess since they were added later, those are the exceptions, not the norm. Alright, we can go with `--autoCategorize` and I can post a PR later that makes the other options consistent.

src/update-changelog.ts
jake-perkins: Updated to Enum to recommended

mcmire:
Looks like there are some lint failures to fix, and then I can approve.

src/update-changelog.ts
jake-perkins: Updated to be optional!

src/update-changelog.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    const result = await ChangeLogManager.updateChangelog({
+      changelogContent: uncategorizedChangelog,
+      currentVersion: '1.0.0',
+      repoUrl:
+        'https://github.com/ExampleUsernameOrOrganization/ExampleRepository',
+      isReleaseCandidate: true,
+      autoCategorize: true,
+    });
+
+    console.log(result);
```
cryptodev-2s, line 32: Nit: this console log can be removed

jake-perkins: Sorry am troubleshooting some unit tests that are passing locally but failing on the pipeline, so these are just temporary will remove

mcmire (approved):
Thanks for taking care of this! LGTM.

PR #226, opened 2025-03-05
Use ts-bridge instead of tsc; restrict imports

mcmire:
- Use `ts-bridge` instead of `tsc` to generate a JavaScript build.
  - This now produces two variants to cater to CommonJS-only and ESM-compatible bundlers.
- Define exports in `package.json`
  - **BREAKING:** Custom, on-demand subpath imports are no longer supported.

## Manual testing steps

- Run `yarn build` and inspect `dist`. You should see `.cjs`, `.mjs`, `.d.cts`, and `.d.mts` files for each TypeScript file in `src/`.
- Run `yarn test`. This now runs a second command, `attw --pack`, which verifies that the built files are correct. This should pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,21 +7,35 @@
     "url": "https://github.com/MetaMask/auto-changelog.git"
   },
   "license": "(MIT OR Apache-2.0)",
-  "main": "dist/index.js",
-  "bin": "dist/cli.js",
+  "sideEffects": false,
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/index.d.mts",
+        "default": "./dist/index.mjs"
+      },
+      "require": {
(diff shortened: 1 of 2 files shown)


mcmire:
@SocketSecurity ignore npm/commander@10.0.1
@SocketSecurity ignore npm/mz@2.7.0
@SocketSecurity ignore npm/validate-npm-package-name@5.0.1

`commander` is used by `@arethetypeswrong/cli`. This package is a wrapper around `child_process`, so shell access makes sense.

`mz` is a transitive dependency of `@arethetypeswrong/cli`. It wraps `child_process`, `fs`, and other Node libraries to provide promisified versions. So shell access makes sense here too.

New author for `validate-npm-package-name` is OK, `lukekarrys` is a known member of the JS community.

PR #227, opened 2025-03-05
Sync supported Node versions w/ module template

mcmire:
- Bump support for Node 18.x to at least 18.20
- Limit support for Node 20.x to at least 20.17
- Add support for Node 22 and higher


cryptodev-2s (approved):
LGTM!

PR #235, opened 2025-03-31
Fix path to `bin` file

Mrtenz:
`bin` was pointing to `dist/cli.js` which no longer exists after #226, causing the CLI to break:

```
$ yarn changelog:validate
node:internal/modules/cjs/loader:1228
  throw err;
  ^

Error: Cannot find module './node_modules/@metamask/auto-changelog/dist/cli.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1225:15)
    at Module._load (node:internal/modules/cjs/loader:1051:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:173:12)
    at node:internal/main/run_main_module:28:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "main": "./dist/index.cjs",
   "module": "./dist/index.mjs",
   "types": "./dist/index.d.cts",
-  "bin": "dist/cli.js",
+  "bin": "dist/cli.mjs",
   "files": [
     "dist"
   ],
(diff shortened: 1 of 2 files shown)


cryptodev-2s (approved):
LGTM!

PR #240, opened 2025-04-25
fix: allow category extraction when scope is used

NicolasMassart:
This PR changes the category extraction to use a RegExp to matches [**Conventional Commits** that start with `feat:` or `fix:` (optionally with a scope like `feat(scope):`)](https://www.conventionalcommits.org/en/v1.0.0/#:~:text=A%20scope%20may%20be%20provided%20to%20a%20commit%E2%80%99s%20type%2C%20to%20provide%20additional%20contextual%20information%20and%20is%20contained%20within%20parenthesis%2C%20e.g.%2C%20feat(parser)%3A%20add%20ability%20to%20parse%20arrays.).

- adds tests for getCategory to make sure all previous behaviours are preserved and new one works.
- exports getCategory to make it testable.
- updates getCategory to use regex to extract known category even when scope is present.

### Regex breakdown

- `^` – Start of the string
- `(feat|fix)` – Must start with either `feat` or `fix`
- `(?:\([^)]*\))?` – Optionally match a scope in parentheses like `(bridge)` or `(ramps)`
- `\s*:\s*` – Requires a colon `:` (with optional spaces before/after)
- `/u` – Unicode flag (as GitHub allows PR titles to contain unicode)

See https://consensys.slack.com/archives/C08N7NQRBPU/p1745578034343929


sethkfman (dismissed):
LGTM

src/update-changelog.ts
```diff
@@ -213,10 +213,12 @@ async function runCommand(command: string, args: string[]): Promise<string[]> {
...
-function getCategory(description: string): ChangeCategory {
-  // Check if description contains a colon
-  if (description.includes(':')) {
-    const [prefix] = description.split(':').map((part) => part.trim());
+export function getCategory(description: string): ChangeCategory {
+  const conventionalCommitPattern = /^(feat|fix)(?:\([^)]*\))?\s*:\s*/u;
+  const match = description.match(conventionalCommitPattern);
+
+  if (match) {
+    const prefix = match ? match[1] : undefined;
```
MajorLift, line 221: Since we already check for `match` being nullish in the previous line, perhaps this should be:

```suggestion
    const prefix = match.length ? match[1] : undefined;
```

src/update-changelog.ts
```diff
@@ -213,10 +213,12 @@ async function runCommand(command: string, args: string[]): Promise<string[]> {
  * @param description - The commit message description.
  * @returns The category of the change.
  */
-function getCategory(description: string): ChangeCategory {
-  // Check if description contains a colon
-  if (description.includes(':')) {
-    const [prefix] = description.split(':').map((part) => part.trim());
+export function getCategory(description: string): ChangeCategory {
+  const conventionalCommitPattern = /^(feat|fix)(?:\([^)]*\))?\s*:\s*/u;
```
MajorLift, line 217: Is there a chance that the section following `^(feat|fix)` gets parsed as a non-capturing group i.e. `(?:pattern)?`?

MajorLift: To prevent this, we could escape the enclosing parentheses:

```suggestion
  const conventionalCommitPattern = /^(feat|fix)\(?:\([^)]*\)\)?\s*:\s*/u;
```

MajorLift: Ah I can see that I misinterpreted your intention here. Please disregard :)

MajorLift, line 217: nit: An empty scope string should probably be considered to be invalid.

```suggestion
  const conventionalCommitPattern = /^(feat|fix)(?:\([^)]+\))?\s*:\s*/u;
```

MajorLift (dismissed):
We should get codeowner approval from @MetaMask/wallet-framework-engineers

cryptodev-2s (approved):
LGTM!

MajorLift (approved):

Gudahtt (approved):
LGTM!

PR #275, opened 2026-02-23
Throw if release section is empty

mcmire:
Currently, a release section is valid if it is empty, i.e., has no changelog entries. This commit reverses that behavior.

---

> [!NOTE]
> **Medium Risk**
> Behavioral change to `validateChangelog` that can newly fail existing changelogs/CI if they contain header-only releases; otherwise limited to validation logic.
> 
> **Overview**
> `validateChangelog` now enforces that **every release section contains at least one changelog entry**, throwing a new `EmptyReleaseError` when a release has zero categorized items.
> 
> Tests are updated to expect this failure for empty releases (both RC and non-RC scenarios), and the `CHANGELOG.md` notes this as a **breaking** change.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f20d3fcc10148ba7ea9001f2fae0e44608a6fe02. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
+### Changed
 
-- Add `--noEmptyReleases` option to `validate` command and `validateChangelog` function which will throw if any release sections are empty ([#275](https://github.com/MetaMask/auto-changelog/pull/275))
+- **BREAKING:** Throw error if any release section has no changelog entries ([#275](https://github.com/MetaMask/auto-changelog/pull/275))
 
 ## [5.3.2]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `--noEmptyReleases` option to `validate` command and `validateChangelog` function which will throw if any release sections are empty ([#273](https://github.com/MetaMask/auto-changelog/pull/273))
+
 ## [5.3.2]
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add `--noEmptyReleases` option to `validate` command and `validateChangelog` function which will throw if any release sections are empty ([#273](https://github.com/MetaMask/auto-changelog/pull/273))
+- Add `--noEmptyReleases` option to `validate` command and `validateChangelog` function which will throw if any release sections are empty ([#275](https://github.com/MetaMask/auto-changelog/pull/275))
 
 ## [5.3.2]
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -183,11 +183,6 @@ type ValidateOptions = {
    * associated pull requests within the repository (true) or not (false).
    */
   ensureValidPrLinksPresent: boolean;
-  /**
-   * Whether to validate that each release section has one or more changelog
-   * entries (true) or not (false).
-   */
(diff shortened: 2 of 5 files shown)


src/validate-changelog.ts
```diff
@@ -23,6 +23,15 @@ export class UncategorizedChangesError extends InvalidChangelogError {
   }
 }
 
+/**
+ * Indicates that a release has no changelog entries.
+ */
+export class EmptyReleaseError extends InvalidChangelogError {
```
cryptodev-2s, line 29: Nit: Do we need to export this Error at root level similar to https://github.com/MetaMask/auto-changelog/blob/main/src/index.ts#L6 ?

src/cli.ts
```diff
@@ -183,6 +183,11 @@ type ValidateOptions = {
    * associated pull requests within the repository (true) or not (false).
    */
   ensureValidPrLinksPresent: boolean;
+  /**
+   * Whether to validate that each release section has one or more changelog
+   * entries (true) or not (false).
+   */
+  noEmptyReleases: boolean;
```
cryptodev-2s: Nit: what do you think about renaming this to `rejectEmptyReleases` or `ensureNonEmptyReleases` to use similar tone of other options ?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `--noEmptyReleases` option to `validate` command and `validateChangelog` function which will throw if any release sections are empty ([#273](https://github.com/MetaMask/auto-changelog/pull/273))
```
cryptodev-2s: Nit: referencing wrong PR
```suggestion
- Add `--noEmptyReleases` option to `validate` command and `validateChangelog` function which will throw if any release sections are empty ([#275](https://github.com/MetaMask/auto-changelog/pull/275))

File: src/cli.ts [hunk 4]
```diff
@@ -403,6 +412,12 @@ async function main() {
               'Verify that each changelog entry has one or more links to associated pull requests within the repository',
             type: 'boolean',
           })
+          .option('noEmptyReleases', {
+            default: false,
```
FrederikBolding: Is there a reason to not enable this functionality by default?

src/validate-changelog.ts
mcmire: I don't see a need for it now. If we need access to this type elsewhere I feel like we can add it later.

src/cli.ts
mcmire: We could, but it would be a breaking change, since any changelogs that already have empty release sections would fail validation now. So I was a bit hesitant to make that change now. That said, I suspect that most projects' changelogs are compliant and would require no work to upgrade. What do you think?

src/cli.ts
mcmire: Good callout. This does seem a bit off. How does `--disallowEmptyReleases` sound?

CHANGELOG.md
mcmire: Oops, thank you!

src/cli.ts
cryptodev-2s: Yes sounds good to me!

src/cli.ts
FrederikBolding: IMO worth it to make the breaking change and deal with the limited invalid changelogs that will come up.

mcmire: Okay, I've made this behavior the default here: [f20d3fc](https://github.com/MetaMask/auto-changelog/pull/275/commits/f20d3fcc10148ba7ea9001f2fae0e44608a6fe02)

src/cli.ts
mcmire: I ended up removing the option entirely: [f20d3fc](https://github.com/MetaMask/auto-changelog/pull/275/commits/f20d3fcc10148ba7ea9001f2fae0e44608a6fe02)

cryptodev-2s (approved):
LGTM!

FrederikBolding (approved):

PR #292, opened 2026-08-05
feat: much better changelog parsing

HowardBraham:
This is absolutely vibe-coded, but heavily tested against 4 weeks of changelogs in both Extension and Mobile..

The following is an AI summary of all the fixes made:

## Changelog entry parsing (`get-new-changes.ts`)

- **Convert leading verb to past tense.** Normalize a changelog entry's
  leading verb to past tense (e.g. `add` → `Added`), driven by a
  `leadingVerbToPastTense` map, with guards to avoid mis-converting
  punctuation, noun phrases, ambiguous words, and coordinated clauses. The
  map is a conservative allow-list; verbs observed in real changelogs but
  missing from it (`store`, `centralize`, `merge`, `reorder`) were added so
  entries like "Store money account results" become "Stored ...". The verb
  `instrument` was deliberately left out: it is more commonly a noun than a
  verb, so auto-converting it risks mangling noun-phrase entries.
- **Multi-entry support with semicolon join.** When a PR body contains two,
  three, or more `CHANGELOG entry:` lines, collect all of them and join the
  values with `; ` instead of keeping only the first.
- **Ignore no-changelog and empty entries when other entries are present.**
  Drop empty and no-changelog entry lines from the joined result, but exclude
  the commit entirely when _every_ entry is a no-changelog opt-out. The
  opt-out is no longer limited to the word `null`: it also recognizes common
  placeholders authors write to mean "no changelog needed", such as `N/A`,
  `TBD`, and `none`. This stops a stray line like `- N/A (#12345)` from leaking
  into the changelog.
- **Ignore `CHANGELOG entry:` lines inside HTML comments.** The pull request
  template includes a commented-out `CHANGELOG entry: null` alternative inside
  an HTML comment (`<!-- ... -->`). Skip anything inside such comment blocks so
  the commented-out line — and its closing `-->` — are never parsed as a real
  entry. This stops a real entry from being joined with junk like
  `; null -->`.
- **Hard-wrapped entry continuation with URL exemption.** Join a
  hard-wrapped continuation line onto its entry, including when the
  continuation line begins with a URL — but only when the entry already has
  text, so an empty entry followed by a URL still falls back to the PR
  subject.
- **Hard-wrapped Markdown link labels.** Preserve a blank line inserted
  inside a Markdown link label by GitHub's commit-message wrapping, such as
  `[patch for core\n\nclient-utils](...)`. Previously the blank line ended the
  entry, truncating the link and causing a release-automation cherry-pick to
  have a different description from its original commit. That prevented
  cherry-pick de-duplication and emitted two malformed entries. A blank line
  is now retained only while the entry has an unmatched `[`; ordinary blank
  lines remain entry boundaries.
- **Standalone Markdown-link entries.** When the entire authored entry is a
  Markdown link, use its readable label rather than the link itself. A
  changelog entry such as `[Patch missing data](https://example.com/pr/1)` is
  now rendered as `Patch missing data`, while Markdown links embedded in a
  longer sentence remain unchanged.
- **CRLF normalization.** Normalize `\r\n` / `\r` to `\n` before parsing entry
  bodies.
- **Drop release-automation cherry-picks.** When a change is cherry-picked
  onto an older release branch and then merged forward, it appears twice in
  the next release's commit range — once from the original commit (which
  targets `main`) and once from the cherry-pick. Both carry the same
  changelog entry but different PR numbers, so the change was listed twice
  (e.g. `Added decimal validation (#44602)` and again `(#44604)`). Detect the
  cherry-pick from its commit subject and, when the original commit is present
  in the same batch, drop the cherry-pick so only the original PR is listed.
  The detection covers the subject forms produced by MetaMask's "Runway"
  release automation across both repositories — `release(runway): cherry-pick`
  (extension), `chore(runway): cherry-pick` (mobile), `release(cp):`, and the
  freeform `Cherry-picking commits from main to release/...` helper. Distinct
  PRs that merely share an identical short entry are left untouched, because
  the drop only applies to commits that are themselves marked as cherry-picks.
  A manual cherry-pick that carries none of these markers is not detected and
  would still need to be de-duplicated by hand; this was judged an acceptable
  trade-off to avoid collapsing genuinely distinct same-titled PRs.

## Categorization (`update-changelog.ts`, `constants.ts`)

- **Keyword-based exclusion.** Exclude commits whose description matches
  configured exclusion keywords (word-boundary matched), including a bare
  `oidc` keyword.
- **Respect author-provided entries for `chore:` commits.** When a
  description came from an explicit `CHANGELOG entry:`, categorize a
  `chore:`-prefixed change by its leading verb instead of excluding it
  (`getCategory`'s `hasChangelogEntry` flag).

## Project detection from `package.json` (`repo.ts`, `cli.ts`)

- **Infer the repository URL from `package.json` when run directly.**
  Previously the repository URL was only auto-detected when the tool was run
  through an npm/yarn script (which sets a helper environment variable), so
  running the CLI directly forced you to pass `--repo`. It now also falls back
  to reading the `repository` field of the nearest `package.json` at or above
  the current working directory, so `--repo` is no longer required in a normal
  project checkout.
- **Infer the current version from the nearest `package.json`.** When
  `--currentVersion` is not provided (and no `--root` is given), the tool now
  finds the nearest `package.json` at or above the current working directory
  and reads its `version` field, rather than only looking at a `package.json`
  directly in the current directory. This matches the repository-URL lookup
  above, so running the CLI from a subdirectory no longer requires
  `--currentVersion`.

## GitHub authentication (`get-new-changes.ts`)

- **Fall back to the GitHub CLI for a token.** When fetching pull request
  labels, the tool still prefers the `GITHUB_TOKEN` environment variable. If
  that variable is not set and the tool is not running on CI, it now falls
  back to the token from the locally authenticated GitHub CLI
  (`gh auth token`). This means developers who have already run `gh auth login`
  no longer have to export a token by hand. On CI (or when `gh` is unavailable
  or unauthenticated) the behavior is unchanged: a missing `GITHUB_TOKEN`
  still errors.

## UPDATE AUGUST 11

- If there's a cherry-picked commit, use the PR number from main, not from the cherry-pick
- Do not list the cherry-picked commit twice
- New cli flags:
  - `--normalizeToPastTense`
  - `--excludeChoreWithoutChangelogEntry`
  - `--verbose`
  - `--preventBackfill` (if a human or AI editor deliberately removes a commit, we want it to stay removed if auto-changelog is run again)

---

> [!NOTE]
> **Medium Risk**
> Large behavioral changes to what commits appear in changelogs and how entries are worded; wrong parsing or dedup could omit or duplicate user-facing release notes across Extension/Mobile workflows.
> 
> **Overview**
> Overhauls how `update` turns git commits into changelog bullets for MetaMask-style PRs, with new CLI switches and much broader test coverage.
> 
> **Commit → entry pipeline** rewrites `CHANGELOG entry:` parsing: multiple entries join with `; `, hard-wrapped bodies and CRLF are handled, HTML-comment template lines and opt-out placeholders (`null`, `N/A`, `TBD`, etc.) are respected, and descriptions get optional **past-tense leading verbs**, trailing-period cleanup, and smarter subject fallbacks. **Runway / release cherry-picks** are detected so the original PR number wins and duplicate lines are dropped when the main commit is in the same batch.
> 
> **Filtering & categorization** adds optional **`--excludeChoreWithoutChangelogEntry`**, **`--preventBackfill`** (skip commits older than PRs already in the target section via git ancestry), and **`--verbose`** stderr diagnostics. Auto-categorize gains leading-verb maps, word-boundary exclusion keywords (e.g. `oidc`), and chore handling when an authored entry exists.
> 
> **DX**: **`findNearestPackageJson`** resolves repo URL and version from cwd; local runs can use **`gh auth token`** when `GITHUB_TOKEN` is unset (not on CI). README documents the new flags; tests use **nock** for GitHub label calls.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 296a590955cfec99b698b1b08e7762a69ae08c11. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tommasini (dismissed):

src/get-new-changes.test.ts
```diff
@@ -1,8 +1,20 @@
+import { Octokit } from '@octokit/rest';
+
 import { getNewChangeEntries } from './get-new-changes';
 import { runCommand, runCommandAndSplit } from './run-command';
 
 jest.mock('./run-command');
 
+// Mock Octokit so the `useChangelogEntry` path (which fetches PR labels) does
```
mcmire: Instead of mocking Octokit what are your thoughts on using Nock? This is what we conventionally use elsewhere to mock requests.

src/get-new-changes.test.ts
```diff
@@ -1,8 +1,20 @@
...
+// not make network calls. The Jest config has `resetMocks: true`, so the
+// constructor implementation is re-established in `beforeEach` below.
+const mockPullsGet = jest.fn();
+jest.mock('@octokit/rest', () => ({
+  Octokit: jest.fn(),
+}));
+
+const mockOctokit = Octokit as unknown as jest.Mock;
```
mcmire: Nit: If we still want to mock Octokit, `jest.mocked` offers a more idiomatic way to work with a mocked import:

``` suggestion
const mockOctokit = jest.mocked(Octokit);
```

We also might want to use a noun form rather than a verb form:

``` suggestion
const OctokitMock = jest.mocked(Octokit);
```

src/constants.ts
```diff
@@ -105,4 +105,143 @@ export const keywordsToIndicateExcluded: string[] = [
...
+ * Maps a lowercase leading verb of a "Keep a Changelog"-style entry to its
+ * category. The MetaMask PR template asks authors to write user-facing entries
+ * in the past tense (e.g. "Added a new tab...", "Fixed a bug..."), so when a
+ * commit has no usable Conventional Commit prefix we can still categorize the
+ * entry deterministically from its first word.
+ *
+ * Only unambiguous, closed-set verbs are included; entries that do not begin
+ * with one of these words remain uncategorized rather than being guessed.
+ */
+export const changelogVerbToCategory: Record<string, ChangeCategory> = {
```
mcmire: Nit: Do we want to use `SCREAMING_SNAKE_CASE` for this since it's a constant?

``` suggestion
export const CHANGELOG_VERB_TO_CATEGORY: Record<string, ChangeCategory> = {
```

src/constants.ts
```diff
@@ -105,4 +105,143 @@ export const keywordsToIndicateExcluded: string[] = [
...
+ * The set is intentionally closed and conservative. Words whose first token is
+ * commonly a noun as well as a verb (e.g. "support", "display", "show",
+ * "swap", "set", "use", "filter", "change", "handle") are deliberately
+ * omitted: converting them risks producing garbage such as "Supported for
+ * Solana snaps". Unknown leading words are left untouched.
+ *
+ * The value is the capitalized past form, since descriptions have already been
+ * capitalized by the time conversion runs.
+ */
+export const leadingVerbToPastTense: Record<string, string> = {
```
mcmire: Similar as above:

``` suggestion
export const LEADING_VERB_TO_PAST_TENSE: Record<string, string> = {

File: src/get-new-changes.ts [hunk 5]
```diff
@@ -78,11 +313,297 @@ function stripHtmlComments(message: string): string {
...
+    // Join multiple entries with `; ` so 2, 3, or more distinct entries are
+    // all preserved in the rendered changelog line. A real entry always wins
+    // over a sibling `null` entry.
+    return entries.join('; ');
+  }
+
+  if (sawNullEntry) {
+    // Every entry was `null` -> the author opted out; return the literal
+    // `'null'` so the commit is excluded downstream.
+    return 'null';
```
mcmire: What do you think about having this function return a result object — something with a `status` or `type` property? Perhaps this case could be represented as `{ status: 'everyEntryNull' }`. It seems strange otherwise that we are using a literal string for this purpose, particularly as `'null'` will collapse into `string` due to the return type of this function.

src/constants.ts
```diff
@@ -105,4 +105,143 @@ export const keywordsToIndicateExcluded: string[] = [
...
+};
+
+/**
+ * Maps a lowercase leading verb (imperative or third-person present) of a
+ * changelog entry to its past-tense form.
+ *
+ * The MetaMask PR template asks for past-tense, user-facing entries, but many
+ * entries fall back to a commit subject written in the imperative mood
+ * ("Add ...", "Fix ...", "Migrate ..."). This map lets us convert only the
+ * leading verb to past tense so the rendered changelog reads consistently.
```
mcmire: Should we clarify that this is in the client PR templates? We don't have this restriction for libraries or other kinds of repos. 

``` suggestion
 * For MetaMask client repos, PR templates asks for past-tense, user-facing
 * entries, but many entries fall back to a commit subject written in the
 * imperative mood ("Add ...", "Fix ...", "Migrate ..."). This map lets us
 * convert only the leading verb to past tense so the rendered changelog reads
 * consistently.
```

src/constants.ts
```diff
@@ -105,4 +105,143 @@ export const keywordsToIndicateExcluded: string[] = [
...
+ *
+ * Design constraints.
+ *
```
mcmire: Was this left in accidentally?

``` suggestion
```

src/constants.ts
```diff
@@ -105,4 +105,143 @@ export const keywordsToIndicateExcluded: string[] = [
...
+ * Every past form is spelled out explicitly (no generative morphology), so
+ * irregular verbs and doubled-consonant spellings are always correct.
+ *
+ * The set is intentionally closed and conservative. Words whose first token is
```
mcmire, line 159: I know you said you tested this for a while, but I am curious how frequently we will need to go back and add new words in the future. Have you considered using a denylist rather than an allowlist? Or would that require the same amount of work?

src/get-new-changes.ts
```diff
@@ -218,24 +761,80 @@ function deduplicateCommits(
...
+ *
+ * @param description - The raw commit or changelog-entry description.
+ * @returns The normalized description.
+ */
+function normalizeDescriptionForChangelog(description: string): string {
+  const [firstLine, ...remainingLines] = description.split('\n');
+  const normalizedEntries = firstLine
+    .split('; ')
+    .map((entry) =>
+      convertLeadingVerbToPastTense(normalizeTrailingPeriod(entry)),
```
mcmire: It looks like we always run through this conversion step when copying commit messages to changelogs. What do you think about putting this behind a flag? I feel like the clients are the only place where we'd want to do this. For libraries and other kinds of projects, we have a soft convention that changelog entries are written in imperative tense (i.e. it's not written down but most people tend to do it instinctively).

src/get-new-changes.ts
```diff
@@ -287,48 +886,103 @@ export async function getNewChangeEntries({
...
-    return { description: newDescription, subject: subjectForCategorization };
-  });
+      if (prNumber) {
+        const suffix = useShortPrLink
+          ? `(#${prNumber})`
+          : `([#${prNumber}](${repoUrl}/pull/${prNumber}))`;
+
+        if (newDescription) {
+          const lines = newDescription.split('\n');
+          lines[0] = `${lines[0]} ${suffix}`; // Append suffix to the first line (next lines are considered part of the description and ignored by the parsing logic)
```
mcmire: Nit: Perhaps this comment should go on its own line for readability?

``` suggestion
          // Append suffix to the first line (next lines are considered part of
          // the description and ignored by the parsing logic)
          lines[0] = `${lines[0]} ${suffix}`;

File: src/get-new-changes.ts [hunk 11]
```diff
@@ -287,48 +886,103 @@ export async function getNewChangeEntries({
...
+        }
+      }
+
+      return {
+        description: newDescription,
+        subject: subjectForCategorization,
+        // Whether a usable `CHANGELOG entry:` was authored for this commit.
+        // Used downstream to exclude category-less CI commits (e.g. `chore:`)
+        // that carry no user-facing entry.
+        hasChangelogEntry: changelogEntry !== undefined,
```
mcmire, line 1189: Maybe we can make this decision in `getCommits` and move this property to the `Commit` object? Then we wouldn't need to explain this here.

src/get-new-changes.ts
```diff
@@ -100,7 +621,7 @@ async function getCommits(
 ): Promise<Commit[]> {
   // Only initialize Octokit if we need to fetch PR labels
   if (useChangelogEntry) {
-    initOctoKit();
+    await initOctoKit();
```
mcmire, line 665: I know we are changing pre-existing code, but why do we have to conditionally initialize Octokit? When we say `new Octokit`, does it make requests or something?

src/get-new-changes.ts
```diff
@@ -348,7 +1002,7 @@ async function getPrLabels(
   }
 
   if (!github) {
-    initOctoKit();
+    await initOctoKit();
```
mcmire, line 1254: Similar as above, why do we have to conditionally initialize Octokit?

src/repo.test.ts
```diff
@@ -2,9 +2,27 @@
...
+      delete process.env.npm_package_repository_url;
+    } else {
+      process.env.npm_package_repository_url = originalNpmPackageRepositoryUrl;
+    }
+    if (originalProjectCwd === undefined) {
+      delete process.env.PROJECT_CWD;
+    } else {
+      process.env.PROJECT_CWD = originalProjectCwd;
+    }
+    jest.restoreAllMocks();
```
mcmire: This step shouldn't be necessary, as we already have `restoreMocks: true` defined in `jest.config.js`.

``` suggestion
```

src/repo.ts
```diff
@@ -26,18 +88,20 @@ export function getRepositoryUrl(): string | null {
...
+    );
+    if (repositoryUrl) {
+      return repositoryUrl;
     }
+  }
 
-    if (typeof packageJsonContent.repository?.url === 'string') {
-      return packageJsonContent.repository.url.replace(/\.git$/u, '');
-    }
+  // When run directly (not through an npm/yarn script) neither env var above
```
mcmire, line 99: Was this added to support the `yarn tsx ../auto-changelog ...` workflow you mentioned earlier?

src/update-changelog.ts
```diff
@@ -272,22 +294,85 @@ export function getCategory(description: string): ChangeCategory {
       case ConventionalCommitType.RELEASE:
         return ChangeCategory.Excluded;
       // End categories that should be excluded from the changelog
+      // A `chore:` commit with no authored `CHANGELOG entry:` is a CI/internal
```
mcmire, line 345: Should we put this behind a flag as well? For libraries, `chore:` doesn't always mean that it's a developer-only change; it's often used as a catch-all ([for instance](https://github.com/MetaMask/core/commit/85fce1ea98ea5d91e529b9c1baa3f635a528c284)).

mcmire (commented):
For the most part these changes seem innocuous, but I did call a couple that could be surprising when used in `core` or other repos.

src/repo.ts
HowardBraham: Yep, it's so I could test this faster without publishing or linking.

src/get-new-changes.ts
HowardBraham: Implemented as `--normalizeToPastTense`

src/update-changelog.ts
HowardBraham: Implemented as `--excludeChoreWithoutChangelogEntry`

src/constants.ts
HowardBraham: I think the next step is actually to run it through this, and then run it through an LLM running on CI.  This could never do everything.

mcmire: Ah yeah that's probably a good plan.

mcmire (approved):
Looks great! Good call on adding `--preventBackfill` too.

